PR #11, opened 2020-03-26
add two grafana dasboards for cost control analysis

verdverm:
1. Airflow deployments. This uses more appropriate metrics for pods and allows selection by deployment.
2. Kubernetes Nodes. Displays the metrics important to understanding request capacity consumption.

### DIFF
--- a/grafana/include/airflow-utilization.json
+++ b/grafana/include/airflow-utilization.json
@@ -1,0 +1,2502 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": "-- Grafana --",
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "id": 26,
+  "iteration": 1578687477848,
+  "links": [],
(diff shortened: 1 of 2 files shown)


schnie:
@verdverm these are generic enough that it makes sense to ship to enterprise customers right?

Do you have a screenshot or link for each of these?

verdverm:
The kubernetes nodes queries are specific to prod cloud, need a way to discover the pools and setup two panels per pool.

Mostly wanted to get these somewhere besides my laptop so we can iron out these details and if this is the appropriate repo to host them.

verdverm:
Screenshot from the Kubernetes Node Dashboard

Screenshot from 2020-03-26 08-00-52

Currently the Airflow Deployment is not in prod cloud, but maybe old cloud? (if it hasn't been  redeployed since I put it there)

verdverm:
@schnie You said astro specific dashboards should go in "Velero?" Where is that?

I was also thinking that we could have a generic version of the node dash if we just lump all nodes into one chart. This would probably be useful to the operators running a single node pool.

sjmiller609:
> @schnie You said astro specific dashboards should go in "Velero?" Where is that?
> 
> I was also thinking that we could have a generic version of the node dash if we just lump all nodes into one chart. This would probably be useful to the operators running a single node pool.

specific dashboards go here https://github.com/astronomer/astronomer/tree/master/charts/grafana/dashboards

that is disabled by default

ryw:
is this good to merge or needs more work?

PR #13, opened 2020-04-01
Add prometheus postgres exporter image

sjmiller609:

https://github.com/astronomer/issues/issues/950

**Summary of changes**:

The purpose of this change is to support adding in a new prometheus exporter, this one is designed to serve metrics based on the content of a postgresql database. This software supports adding in new queries that can produce new metrics.

This is the software https://github.com/wrouesnel/postgres_exporter

- [x] postgres-exporter

**Checklist** (required)

- [x]  an [issue](https://github.com/astronomer/issues/issues/1082) has been created 

If adding a new image:

- [x] the directory has the same name you intend it to be published as, less a preceding "ap-"
- [x] the directory includes a file "version.txt"
- [x] the directory includes a file "cve-whitelist.yaml", which may be empty
- [x] the file .github/PULL_REQUEST_TEMPLATE.md is updated to include the image in the checklist
- [x] execute the script .circleci/generate_circleci_config.py, commit changes to .circleci/config.yml

@schnie you have to do the adding of new image since CI does not have permissions


schnie:
Interesting, looks useful. Whats the motivation right now behind this? Are you planning a helm PR with it?

sjmiller609:
yep, helm PR will come. it's to alert on the DB in general, trigger by the
issue around 'version' in the 'Deployment' table

sjmiller609:
I made an issue for the vulnerabilities on ES vendored products

postgres-exporter/Dockerfile
```diff
@@ -0,0 +1,40 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# https://github.com/wrouesnel/postgres_exporter/blob/v0.8.0/Dockerfile
+FROM wrouesnel/postgres_exporter:v0.8.0 AS upstream
+# The above 'upstream' layer is FROM scratch (they use a multi-stage build)
+# I would like some normal tooling like package manager and shell available.
```
jpweber, line 19: I respectfully disagree with this. Containers should _not_ have any extra stuff in them that they don't need. If this is from scratch that is already ideal for a go binary.

postgres-exporter/include/entrypoint
```diff
@@ -0,0 +1,18 @@
+#!/bin/sh
+
+function append_ssl_mode {
```
jpweber, line 3: what do we need this for? If I understand how this works correctly it just uses what is in the ENV var for `DATA_SOURCE_NAME` shouldn't that have that correct value ssl or not?

postgres-exporter/Dockerfile
sjmiller609: then how ought we to debug DB connection issue if we cannot exec into pod

postgres-exporter/include/entrypoint
sjmiller609: because the connection string does not always include the SSL option. for example in dev cloud the secret does not include the ?sslmode=disable argument on the connection string, but still the DB does not have SSL enabled

sjmiller609: We discussed in Zoom just now and decided let's add a comment here on the difference between DATA_SOURCE_NAME vs DATA_SOURCE_URI.

postgres-exporter/include/healthcheck
```diff
@@ -0,0 +1,9 @@
+#!/bin/sh
+set -xe
+PORT=9187
+if ! [[ -z "$1" ]]; then
+  PORT=$1
+fi
+# This container is considered unhealthy when it can't
+# connect to the database.
+wget -O - http://localhost:$PORT/metrics | grep 'pg_up 1'
```
sjmiller609, line 9: Jim and I discussed this one, and decided that the metrics server for PG is better to remain running when PG is down. We need to test that if this pod is started without PG running, then PG starts, that the logs indicate that it established a connection (i.e. this program handles PG going down, then coming back up)

PR #19, opened 2020-04-15
bump alpine to 3.11

andriisoldatenko:

- [x] registry

based on https://github.com/docker/distribution-library-image/tree/0b6ea3ba50b65563600a717f07db4cfa6f18f957/amd64


andriisoldatenko:
since docker doesn't react to my old PR https://github.com/docker/distribution-library-image/pull/92

registry/config-example.yml
```diff
@@ -0,0 +1,18 @@
...
+    rootdirectory: /var/lib/registry
+http:
+  addr: :5000
+  headers:
+    X-Content-Type-Options: [nosniff]
+health:
+  storagedriver:
+    enabled: true
+    interval: 10s
+    threshold: 3
```
schnie, line 18: @andriisoldatenko newline here and in the entrypoint?

andriisoldatenko: good catch.
i was wondering instead of adding binary to repo just curl it from github releases?

PR #29, opened 2020-05-27
Resolve CVE-2020-8165 on fluentd

ryw:
Resolves CVE https://github.com/astronomer/ap-vendor/network/alert/fluentd/include/Gemfile/activesupport/open

@andriisoldatenko do you have a way to test this image?

### DIFF
--- a/fluentd/include/Gemfile
+++ b/fluentd/include/Gemfile
@@ -1,6 +1,6 @@
 source 'https://rubygems.org'
 
-gem 'activesupport', '6.0.2.1'
+gem "activesupport", ">= 6.0.3.1"
 gem 'json', '>=2.3.0'
 gem 'rake', '>= 12.3.3'
 gem 'fluentd', '1.9.2'


andriisoldatenko:
@ryw yeah only by deploying and check logs in airflow UI.

ryw:
ugh - we need to develop an isolated test harness for each image that we can run in CI - anyone talking about that? /cc @vishwas-astro

schnie:
We're building a platform test framework. Adding a test to read the logs somehow could work.

ryw:
@schnie don't you think we should build an isolated test though for each image that doesn't require integration w/ Airflow?

schnie:
Yea we could, but maybe a bigger bang for our buck is building it into our up and coming platform suite.

The image may run fine on its own, but say we upgrade and something like the Kubernetes plugin and it behaves differently, which causes our fluentd config to not work properly, which breaks the platform behavior. I'm kind of more worried about something "small" changing within the context of our exact usage, than something big breaking.

Both are probably fine ways to do it, I'm just thinking we could build a more comprehensive / useful test or two up into the platform suite. Depends on whats involved with each I guess.

andriisoldatenko:
@schnie PTAL

PR #32, opened 2020-06-03
Modified airflow-state dashboard to support “ALL”

jpweber:
https://github.com/astronomer/issues/issues/1371

**Summary of changes**:
changed where resource quote label is filtering to use =~ rather than = so regex can be used.

- [ ] alertmanager
- [ ] curator
- [ ] elasticsearch-exporter
- [x] grafana
- [ ] kube-state
- [ ] nginx
- [ ] pgbouncer
- [ ] prisma
- [ ] redis
- [ ] statsd-exporter
- [ ] elasticsearch
- [ ] fluentd
- [ ] kibana
- [ ] kubed
- [ ] nginx-es
- [ ] pgbouncer-exporter
- [ ] prometheus
- [ ] node-exporter
- [ ] registry
- [ ] keda
- [ ] keda-metrics-adapter
- [ ] blackbox-exporter
- [ ] postgres-exporter

**Checklist** (required)

- [ ]  an [issue]() has been created

If adding a new image:


grafana/version.txt
```diff
@@ -1 +1 @@
-7.0.1
+7.0.1-r1
```
schnie, line 1: Looks great, just wondering what we are doing in the other images for the modifier here? 

I remember chatting with @vishwas-astro recently and wondered if we could use `+astro.1` as the modifier to make it clear across the board whats happening.

schnie: Incrementing the number as we go of course.

jpweber: I think I tried that once and it failed in CI because the ci system checks for semver against pep 440. Here are examples that are compliant. https://www.python.org/dev/peps/pep-0440/#examples-of-compliant-version-schemes

it looks like we could maybe to `.astro1` Of course we could always change the semver check too. I'm game for whatever, I just found that -r1 worked. :)

schnie: Ah, ok `.astro1` works for me. Then maybe we can start standardizing on that format across the rest of the vendors when we get to it. Mostly concerned with just getting something consistent that we can roll with.

jpweber: Version has been updated.

jpweber:
Before
Screen Shot 2020-06-03 at 2 31 08 PM

After
Screen Shot 2020-06-03 at 2 29 23 PM

PR #37, opened 2020-07-14
Improve statsd metrics cardinality

sjmiller609:

https://github.com/astronomer/issues/issues/1528

**Summary of changes**:

- [ ] alertmanager
- [ ] curator
- [ ] elasticsearch-exporter
- [ ] grafana
- [ ] kube-state
- [ ] nginx
- [ ] pgbouncer
- [ ] prisma
- [ ] redis
- [x] statsd-exporter
- [ ] elasticsearch
- [ ] fluentd
- [ ] kibana
- [ ] kubed
- [ ] nginx-es
- [ ] pgbouncer-exporter
- [ ] prometheus
- [ ] node-exporter
- [ ] registry
- [ ] keda
- [ ] keda-metrics-adapter
- [ ] blackbox-exporter
- [ ] postgres-exporter
- [ ] nats-streaming

**Checklist** (required)

- [ ]  an [issue]() has been created

If adding a new image:

The objective of this PR is to reduce [cardinality](https://www.robustperception.io/cardinality-is-key) in our Airflow metrics. We ought to not include task or dag ID in any metric, since the space for such a label to too large.


statsd-exporter/include/mappings.yml
```diff
@@ -1,19 +1,20 @@
...
     name: "airflow_operator_successes"
-    labels:
-      operator: "$1"
+    labels: {}
   - match: airflow.operator_failures_(.*)
     match_type: regex
     name: "airflow_operator_failures"
-    labels:
-      operator: "$1"
+    labels: {}
```
ashb, line 17: These ones are used currently:

image

sjmiller609:
I'm going to think a bit to see if I can find a way to add a test for this change in this repo.

statsd-exporter/include/mappings.yml
sjmiller609: @ashb I think we have to sacrifice this feature intentionally. For example, one customer is using custom-named operators like CustomerNamePythonOperator_asdf23213 and there are hundreds of them. This is causing the UI to freeze and prometheus to overload on the backend. :/

ashb: Before we make this change the UI will need to be updated, won't it?

sjmiller609: I happen to have worked with @samblackk on this yesterday. I will instead show:
"Value: x.xxxxx ops"

ashb: Alright, next question: is this a config we only want to apply in our cloud, but for enterprise installs it can stay with the current, more granular, labels?

sjmiller609: That's a good question, and it is up for discussion. My thought is that high cardinality will be seen as a bug by experienced enterprises. In particular, users consuming AC but not enterprise with the helm chart and their own metrics set up I think will desire this behavior.

I think we ought to default to a performant and scalable configuration. In particular dag_id and task_id will very quickly become too large for the average user but operator might not for the average user consuming only normal operators.

ashb: Dag_id+task_id I can see very quickly getting out of hand.

dag_id alone _might_ be okay, and for _most_ deployments, the operator one will be very low cardinality.

sjmiller609: what do you think ash?

sjmiller609:
sidebar: bumping ES to 7.8.0 did not resolve CVE-2020-12049 , I will loop back to this after addressing testing

statsd-exporter/include/mappings.yml
sjmiller609: I agree this should be valid most of the time, and it's a fairly nice to have label in those cases. I'm going to first check for a best of both worlds option like truncating cardinality somehow to see if that's possible (PythonOperator, BashOperator, ...., and Other) or yes-listing a set of operators, then the other fall into "other" or something like that.

sjmiller609:
while reading the config options, I have found this part https://github.com/prometheus/statsd_exporter#mapping-cache-size-and-cache-replacement-policy and that seems very related considering that _incoming_ metrics are high cardinality and we have found memory issues in statsd. I will also consider this part for performance

statsd-exporter/include/mappings.yml
sjmiller609: I'm not finding a config option for this purpose, so i'm going to check on the airflow side to see if this could be done

sjmiller609: it should doable e.g. here https://github.com/astronomer/airflow/blob/dd6aa13a805bbb3e2a180758f5cb572590f37950/airflow/models/taskinstance.py#L1149 and we could also have an optional configuration for allow list in the airflow config. If not specified then any operator name and if specified only a the configured operator names WDYT @ashb, should I proceed down this path of Airflow change?

sjmiller609: https://github.com/apache/airflow/issues/9823

ashb: LGTM for now.

Let's see if we can add some of this granularity back later.

elasticsearch/Dockerfile
```diff
@@ -13,7 +13,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.7.1
+FROM docker.elastic.co/elasticsearch/elasticsearch-oss:7.8.0
```
ashb, line 16: We probably want to move this to a separate PR though, right?

sjmiller609: yep, I think so.

sjmiller609:
I'll merge this after I add a few tests. The scaffolding is working well now

statsd-exporter/test/.gitignore
```diff
@@ -0,0 +1,2 @@
+venv
```
danielhoherd, line 1: The root `.gitignore` contains `venv` which should match this dir and its subdirs. Is there a use case where this is explicitly needed?

schnie: Maybe from testing this script? https://github.com/astronomer/ap-vendor/blob/master/.circleci/generate_circleci_config.py @sjmiller609 would know

danielhoherd: What I mean is the presence of `venv` in the root .gitignore should catch all instances of venv in all subdirs.

https://git-scm.com/docs/gitignore

> When deciding whether to ignore a path, Git normally checks gitignore patterns from multiple sources... Patterns read from a .gitignore file in the same directory as the path, or in any parent directory

sjmiller609: you are right, it's simply unnecessary

PR #70, opened 2021-01-14
Bump fluentd, nginx.

danielhoherd:
## Description

- Bump fluentd to 1.12.0
- Swap out nginx base image for an image in a different registry. It looks like the project quit pushing images to quay and started pushing them to gcr.io. I did a `docker history` on both the old image and the new one and they look very similar.


tommyk (approved):
nothing will use these images automatically right?  IE. someone would have to make a change upstream to "use" one of these new image versions w/ the new software right?

jpweber (approved):

danielhoherd:
> nothing will use these images automatically right? IE. someone would have to make a change upstream to "use" one of these new image versions w/ the new software right?

Correct. This just publishes a version of our build of our repo. We would then have to reference that version number in our product to actually use it.

jpweber:
> nothing will use these images automatically right? IE. someone would have to make a change upstream to "use" one of these new image versions w/ the new software right?

Correct. A helm chart change is required to start using the new images here.

PR #103, opened 2021-06-30
bumping fluentd version (#2937)

pgvishnuram:

**Summary of changes**:

- [ ] alertmanager
- [ ] blackbox-exporter
- [ ] configmap-reloader
- [ ] curator
- [ ] elasticsearch
- [ ] elasticsearch-exporter
- [X] fluentd
- [ ] grafana
- [ ] keda
- [ ] keda-metrics-apiserver
- [ ] kibana
- [ ] kube-state
- [ ] kubed
- [ ] nats-server
- [ ] nats-streaming
- [ ] nginx
- [ ] nginx-es
- [ ] node-exporter
- [ ] pgbouncer
- [ ] pgbouncer-exporter
- [ ] postgres-exporter
- [ ] prisma
- [ ] prometheus
- [ ] redis
- [ ] registry
- [ ] statsd-exporter

**Checklist** (required)

- [X] version.txt is updated in the changed image(s)

- [X]  an [issue]() has been created

https://github.com/astronomer/issues/issues/2937

If adding a new image:


danielhoherd (approved):
Looks good to me. Since we are deviating from just importing the upstream package versions we should probably update the instructions with a simple way to make sure these versions are the latest compatible set.

vishwas-astro (approved):

vishwas-astro:
> Looks good to me. Since we are deviating from just importing the upstream package versions we should probably update the instructions with a simple way to make sure these versions are the latest compatible set.

@danielhoherd - I thought we are just pulling in 1.12.4 of the upstream fluentd package and some of the other dependencies in the gemfile.

If instructions are needed, where do we update them.

danielhoherd:
@vishwas-astro There are [instructions in the Gemfile itself](https://github.com/astronomer/ap-vendor/blob/main/fluentd/include/Gemfile#L1-L2) about how the file used to be updated. In more recent updates we have deviated from that straight copy/paste to a more nuanced version bump. I looked around for a way to auto-update the Gemfile dependencies to the most recent compatible versions but I can't find anything in Ruby that will do that, which is pretty surprising and also disappointing.

vishwas-astro:
ok @danielhoherd. Thanks for the explanation.

Please suggest a way forward. Thanks!

danielhoherd:
What I'd really like is a tool to auto-resolve the most recent compatible package versions. Something like [pip-compile](https://github.com/jazzband/pip-tools) but for Gemfile. Otherwise we'll have to only update the instructions to say something like "Manually resolve the latest package versions"

danielhoherd:
I opened a discussion on a possible new way to do this where we can dynamically update our versions while still pinning them and keeping track of them. I think it's better than what we have right now. https://astronomer.slack.com/archives/C02NZELT2/p1625616354407900

pgvishnuram:
upstream kubernetes still uses fluent-plugin-kubernetes_metadata_filter 2.6.0 version, we have moved to 2.7.0 in our gem file to address reported CVE.

PR #106, opened 2021-07-07
Try fluentd pinning with Gemfile.lock

danielhoherd:
As a follow up to https://github.com/astronomer/ap-vendor/pull/103#issuecomment-875171051 this is a rework of how we build the fluentd image. This uses the Gemfile as a loose list of source packages to include, and uses Gemfile.lock as a list of more specific versions.

https://astronomer.slack.com/archives/C02NZELT2/p1625616354407900


rbankston (approved):
Looks good. Nothing out of the ordinary for a Gemfile.lock commit and version update

PR #107, opened 2021-07-08
2868 do not overwrite tags

danielhoherd:
The main purpose of this branch was to prevent quay.io images from being overwritten. I have solved this. While digging into the problem I reworked the CI process to rely on external scripts vs yaml-embedded scripts. This gives us more confidence that these are syntactically correct and takes steps towards centralizing these, since they are easier to manage as files.

- Move embedded yaml scripts out to per-file scripts in bin directory
- Move circleci config generation script into bin directory
- Improve pre-commit configs and add new hooks
- Improve Makefile helper targets
- Improve wording in various places
- Run all pre-commit hooks on entire repo, which rewrote some python and yaml in some of these projects
- Manually fixed some typos reported by codespell

Related to astronomer/issues#2868

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -687,15 +687,14 @@ jobs:
           name: Validate semver tag
           command: check_version.py << parameters.directory >>
       - push-to-docker-hub:
-          comma_separated_tags: "latest,$TAG"
+          comma_separated_tags: "$TAG"
           image_name: ap-<< parameters.directory >>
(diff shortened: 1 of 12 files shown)


danielhoherd:
The failures here with trivy are not related to changes in this branch. Those are failing in master, and will be handled in a different branch.

bote795 (approved):
lgtm

PR #111, opened 2021-07-12
Delete prisma because we have prisma2

danielhoherd:
Delete prisma build because we use prisma2 now. The upstream prisma builds have not been updated in 2 years.

Solves https://github.com/astronomer/issues/issues/2930


vishwas-astro (approved):

vishwas-astro:
@andriisoldatenko - can we approve and merge this too?

andriisoldatenko (approved):

PR #113, opened 2021-07-14
USER nobody everywhere that USER doesn't exist yet.

danielhoherd:
Use `USER nobody:nobody` anywhere that USER didn't already exist. I was thinking of going with the astro user, but really, what difference does it make? `nobody` is built in so that seemed easier.

I'm not sure what repercussions this will have in our app though so we should find a good way to test this out before merging it. We may want to remove some of these for niche reasons.

Related to https://github.com/astronomer/issues/issues/1290


andriisoldatenko:
@danielhoherd i can test it locally in kind later today.

PR #115, opened 2021-07-16
registry: image should be created with a non root user

andriisoldatenko:
https://github.com/astronomer/issues/issues/1290


andriisoldatenko:
practically LGTM.

danielhoherd (approved):

vishwas-astro (approved):

registry/version.txt
```diff
@@ -1 +1 @@
-3.13.7
+3.13.5-1
```
vishwas-astro, line 1: Why 3.13.5-1 and not 3.13.7-1?

PR #118, opened 2021-07-16
es: image should be created with a non root user

andriisoldatenko:
Related to https://github.com/astronomer/issues/issues/1290


vishwas-astro (approved):

pgvishnuram:
@vishwas-astro - yes this is fine, for kibana we have done the same.

andriisoldatenko:
> @vishwas-astro - yes this is fine, for kibana we have done the same.

yep

i added USER to kibana) https://github.com/astronomer/ap-vendor/pull/34/files
i thought i  did same for ES but sounds like we miss it

andriisoldatenko:
looks like we can't fix it CVE-2020-25649 we can only ignore it :(

pgvishnuram (approved):

vishwas-astro:
@andriisoldatenko - looks like this failed the trivy scan. Do we need to add some CVE to the ignore file?

PR #125, opened 2021-07-23
ap-blackbox-exporter: image should be created with a non root user

andriisoldatenko:
https://github.com/astronomer/issues/issues/1290


blackbox-exporter/Dockerfile
```diff
@@ -13,9 +13,47 @@
...
+RUN set -ex \
+    && apk add --no-cache make git curl
+
+WORKDIR $DISTRIBUTION_DIR
+
+RUN git clone https://github.com/prometheus/blackbox_exporter $DISTRIBUTION_DIR && git checkout v0.19.0
+
+RUN CGO_ENABLED=0 make build
+
+FROM quay.io/astronomer/ap-base:3.13.5
```
vishwas-astro, line 38: Can we go to ap-base:3.14.0?

andriisoldatenko: AFAIK it doesn't exists

danielhoherd: ap-base 3.14.0 does exist
```suggestion
FROM quay.io/astronomer/ap-base:3.14.0
```

danielhoherd (approved):

PR #110, opened 2021-07-12
Bump to ap-base:3.14

danielhoherd:
## Summary of changes

Alpine 3.14 was merged to ap-base in https://github.com/astronomer/ap-base/pull/21 and all the images that reference it have been updated here

## Images are updated by this PR

- [X] curator
- [X] nginx-es
- [X] postgres-exporter
- [X] registry

## Checklist

Place an '[x]' (no spaces) in all applicable fields. Please remove unrelated fields.

- [X] version.txt is updated in the changed image(s)

PR #136, opened 2021-08-17
Bump registry to 3.14

danielhoherd:
This is a follow up to <https://github.com/astronomer/ap-vendor/pull/110>. Registry is the only project that was having a build failure, so I broke it out into this branch. I think we still need to wait for [a new version of cimg/python](https://github.com/astronomer/ap-vendor/pull/110#issuecomment-892072666)


pgvishnuram:
@danielhoherd  until 3.14 gets fixed can we proceed with 3.13 with other fixable CVE's .

danielhoherd:
@pgvishnuram yes, absolutely. We should not wait for 3.14. We do not have a timeline for when that build will succeed, so we cannot hold our breath and wait for it.

pgvishnuram:
@danielhoherd / @andriisoldatenko  After upgrading docker server version from 20.10.2 to 20.10.6 - the build passed successfully
https://app.circleci.com/pipelines/github/astronomer/ap-vendor/629/workflows/1baa306c-23bd-4060-98d9-153292a2e254/jobs/32313

PR #162, opened 2021-10-07
Fix CVE-2021-22945

danielhoherd:
Related to https://github.com/astronomer/issues/issues/3580


nginx/version.txt
```diff
@@ -1 +1 @@
-0.49.0-1
+0.49.3
```
bote795, line 1: why is it .3 instead of .1 or .2?

danielhoherd: Because we're using k8s.gcr.io/ingress-nginx/controller:v0.49.3

bote795: ahh to match versions i s

PR #174, opened 2021-11-17
Vendor k8s.gcr.io/git-sync/git-sync

danielhoherd:
Add new image: git-sync

Related to https://github.com/astronomer/issues/issues/3708


bote795:
dumb question, but why do we need to add it to our own image list as well?

danielhoherd:
@bote795 mirroring third party software is generally a good idea.

- it lets us control the tags, which are not indelible. we'd be in a bind if somebody deleted or overwrote their tags, and we now have the ability to retroactively replace the image behind the tags if we ever needed to.
- it lets us keep a copy where we know it to be, even if they delete or move the project. (EG: tiller moving from gcr.io to ghcr.io)
- it lets us add in any other bits that we need, like metadata or packages. there are none in this case... yet.
- it generally gives us more control over the contents of the image. EG: we were able to swap out the registry image from one that pulls from an upstream repo to one that build registry from source, and we didn't have to change the location of that image.

Honestly we should be doing this with way more packages. Some day we likely will. EG: at past jobs we kept complete repository mirrors for every of distro of linux we used, and only installed from there. Same with python packages.

It's a lot like forking a repo on GH and then working off that repo.

danielhoherd:
@bote795 see also: https://qz.com/646467/how-one-programmer-broke-the-internet-by-deleting-a-tiny-piece-of-code/

pgvishnuram (approved):

danielhoherd:
There's a security scan failure due to having an older base image. I'm not sure how best to solve this. Will open an issue upstream.

danielhoherd:
I'm merging this with the CVE's added as exceptions in order to unblock testing. I've linked the upstream ticket in the trivyignore file.

PR #194, opened 2021-12-27
Bump some CI image versions. Use var for docker version.

danielhoherd:
Minor CI maintenance and improvements. Address <https://github.com/aquasecurity/trivy/discussions/1515>

### DIFF
--- a/bin/trivy-scan.sh
+++ b/bin/trivy-scan.sh
@@ -8,10 +8,11 @@ scan_target="$1"
 
 set +exo pipefail
 
-trivy image \
+trivy \
+  --cache-dir /tmp/workspace/trivy-cache \
+  image \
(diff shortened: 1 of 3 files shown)


bin/trivy-scan.sh
```diff
@@ -9,9 +9,10 @@ scan_target="$1"
 set +exo pipefail
 
 trivy \
+  --cache-dir /tmp/workspace/trivy-cache \
+  image \
```
aliotta, line 13: is this line intentional?

danielhoherd: Yeah. See https://github.com/aquasecurity/trivy/discussions/1515

aliotta (approved):

PR #205, opened 2022-01-07
Try out grafana 8

danielhoherd:
Bump grafana from 7.5.12 to 8.3.3. This is a big jump so it should go through a lot of QA before shipping with astronomer/astronomer.

We also may have to iterate on this build a bit because it includes some file copies and stuff that may not apply or may have different locations in the new version.

I also tweaked the fluentd dashboard to have two additional graphs.

Related to https://github.com/astronomer/issues/issues/3292


pgvishnuram (approved):
Lets try out

PR #207, opened 2022-01-13
Add auth-sidecar image

rob-1126:
## Summary of changes

Add the nginx-unprivileged image that we use for auth sidecar to our repository. This helps us move towards our goal of carrying all images into our repository so that we can have better change control, simplify image mirroring for airgapped customers, and provide security fixes in advance of upstream.

## Images are updated by this PR

+ [ x ] nginx-unprivileged - added

## Checklist

- [ x ] version.txt is created

- [ x] added an apt-get install libgmp10 -y section to address CVE-2021-43618 upstream

If adding a new image:

- [ x ] name seems OK but has been unreviewed, wouldn't object to something like 'auth-sidecar' - might prevent customer complaints "why are there so many nginxes, we already have an nginx, you already have an nginx"
- [ x] the directory includes a file "version.txt", with version matching the underlying software version
- [ x ] the file .github/PULL_REQUEST_TEMPLATE.md is updated to include the image in the checklist
- [ ]  WARNING WILL ROBINSON. I did not see generate_circleci_config.py in my git repo. Is this screw new or recently removed? I had already added the image manually to the config file (which I assume is what this does). execute the script .circleci/generate_circleci_config.py, commit changes to .circleci/config.yml. 

## Additional notes for your reviewer
Can you check my pipeline and make sure there are no obvious errors? I manually tested it through the security test phase but haven't tested pushing images at all. Do I need to?


nginx-unprivileged/version.txt
```diff
@@ -0,0 +1 @@
+1.21.0
```
andriisoldatenko, line 1:
```suggestion
1.21.4
```

andriisoldatenko (approved):

danielhoherd (dismissed):
@rob-1126 why is https://github.com/astronomer/ap-vendor/blob/3ae1ad8cc4/nginx-es/Dockerfile not sufficient?

rob-1126:
> @rob-1126 why is https://github.com/astronomer/ap-vendor/blob/3ae1ad8cc4/nginx-es/Dockerfile not sufficient?
These two images certainly look to be identical to me.
@pritt20 you see anything I'm missing? can we just use nginx-es? 
@danielhoherd It's a bit odd seeing an -es tagged image floating around something that's not elastic but i'm fine with that.

danielhoherd:
> > @rob-1126 why is https://github.com/astronomer/ap-vendor/blob/3ae1ad8cc4/nginx-es/Dockerfile not sufficient?
> These two images certainly look to be identical to me.
> @pritt20 you see anything I'm missing? can we just use nginx-es?
> @danielhoherd It's a bit odd seeing an -es tagged image floating around something that's not elastic but i'm fine with that.

I agree that having the `-es` there is a bit odd. Maybe we should drop that. @pgvishnuram do you know any reason why dropping the `-es` would cause problems?

We could either create a new repo or potentially rename the old repo, depending on how docker hub and quay handle that. I think we'd need to create a new repo, which would be fine IMHO.

rob-1126:
I'm thinking this image should probably be named ap-auth-sidecar and either be a clone or a re-tagging of nginx-es. @danielhoherd Wanna make a final call here?

pgvishnuram:
we can have a generic name - if we want to use the same nginx image some where for our app as base image

pgvishnuram:
ap-nginx we already use it for our ingress  controller image 
ap-nginx-es is used only for elasticsearch 

if we remove es it will create a conflict to ap-nginx image

pgvishnuram:
may be ap-nginx-base or ap-webserver or ap-nginx-nonroot

danielhoherd:
We've discussed breaking apart ap-vendor into different repos. Maybe nginx is a good candidate for that, so we could build several varieties off the same git repo. We could then push those varieties into the same docker repo with different tags too, which would cut down on docker repo management. EG: we could have ap-nginx:nonroot-1.21.4 and ap-nginx:1.21.4. WDYT @andriisoldatenko @rob-1126 @pgvishnuram ?

rob-1126:
> We've discussed breaking apart ap-vendor into different repos. Maybe nginx is a good candidate for that, so we could build several varieties off the same git repo. We could then push those varieties into the same docker repo with different tags too, which would cut down on docker repo management. EG: we could have ap-nginx:nonroot-1.21.4 and ap-nginx:1.21.4. WDYT @andriisoldatenko @rob-1126 @pgvishnuram ?

I favor trudging on with the current ap-vendor status quo for now. I'm OK with either renaming elasticsearch to use a generically named image or in copying and pasting and having to touch them both during security updates a few times a year. In general, I'd expect virtually every image to at some point to happen to need some entrypoint script or doo-dad that could go in a shared image or be shoved into the chart level. I think there's a lot for having an obvious place to put that or being ready to have an image assume it's own identity at the drop of a hat.

Would prefer we not keep multiple images in the same repo - even if they are closely related. Having things that are different have obviously different names is very helpful.

rob-1126:
@danielhoherd What are next steps here?

rob-1126:
Do to the relative silence on what to call this image, I've amended the PR to propose we name this logging sidecar image ap-logging sidecar. It's now an exact copy of the existing nginx-unprivileged image, but in the future may diverge. If it never does, we will just have a few more places to bump the build number on a few times a year for routine security updates.

rob-1126 (commented):
auth-sidecar seems like a better name - what it does instead of what it's made of.

rob-1126:
@danielhoherd Thoughts? Would like to get this into 0.29.

auth-sidecar/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+# limitations under the License.
+
+# https://hub.docker.com/r/nginxinc/nginx-unprivileged/tags
+FROM nginxinc/nginx-unprivileged:1.21.4-alpine
+LABEL maintainer="Astronomer <humans@astronomer.io>"
+
+ARG BUILD_NUMBER=-1
+LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
+LABEL io.astronomer.docker.module="default-backend"
+LABEL io.astronomer.docker.component="default-backend"
```
pgvishnuram, line 23: @danielhoherd - we need change the naming

auth-sidecar/Dockerfile
```diff
@@ -0,0 +1,31 @@
...
+
+# Expose HTTP
+EXPOSE 9200
```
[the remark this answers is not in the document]
rob-1126: @pgvishnuram @danielhoherd It's present in https://github.com/astronomer/ap-vendor/blob/3ae1ad8cc4/nginx-es/Dockerfile and this image is intended to be an identical copy of that basically. Should it be removed there as well?

auth-sidecar/Dockerfile
rob-1126: just blindly changed from default-backend to auth-sidecar, let me know if that wasn't the right thing to do, haven't looked into what these are used for

danielhoherd (approved):
Seems fine to me. The naming makes sense.

danielhoherd:
I created merged this, created the needed docker repos in https://hub.docker.com/r/astronomerinc/ap-auth-sidecar and https://quay.io/astronomer/ap-auth-sidecar and made sure that the push worked. <https://app.circleci.com/pipelines/github/astronomer/ap-vendor/870/workflows/5726d12f-026d-42c1-9426-16d677f96e6c/jobs/33271>

PR #227, opened 2022-03-22
Feature/ap-1118 Adding Test Capability for Docker images

mishah334:
## Which issue this PR fixes

In this PR, we are adding the test capability for the docker images.

Issue: https://github.com/astronomer/issues/issues/1118

## Summary of changes

- Adding PyTest file that execute basic test on docker images. Possible tests cases: Root User Test, Default user check, User config check, HTTP service running check.
- Each docker image will hold the [test.yaml](bin/test.yaml), defining the test configurations. Example [alertmanager/test.yaml](alertmanager/test.yaml)
- Adding [docker-compose](docker-compose.yaml) to facilitate the docker build and run.
- Updating circle-ci config to add the test execution.
- Updating Makefile to add build and test command. That also work from local.

### Exceptions

- Currently below image test execution is not possible because of its nature. We require container to be run in background for execution of test. But below images require some fixed configurations to keep it running.

  - keda
  - keda-metrics-apiserver
  - kube-state
  - kubed

- For below images, ROOT user test is disabled. As it is running with root user and that making test fail.

  - dind-golang
  - fluentd
  - git-daemon
  - grafana
  - redis

## Images are updated by this PR

- [x] alertmanager
- [x] awsesproxy
- [x] blackbox-exporter
- [x] configmap-reloader
- [x] curator
- [x] dind-golang
- [x] elasticsearch
- [x] elasticsearch-exporter
- [x] fluentd
- [x] git-daemon
- [x] git-sync
- [x] grafana
- [x] keda
- [x] keda-metrics-apiserver
- [x] kibana
- [x] kube-state
- [x] kubed
- [x] nats-server
- [x] nats-streaming
- [x] nats-exporter
- [x] nginx
- [x] nginx-es
- [x] node-exporter
- [x] openresty
- [x] pgbouncer
- [x] pgbouncer-exporter
- [x] postgresql
- [x] postgres-exporter
- [x] prometheus
- [x] redis
- [x] registry
- [x] statsd-exporter

## Checklist

- [ ] version.txt is updated in the changed image(s)

- [ ] a fix has been applied OR...
- [ ] an [issue]() has been created

If adding a new image:


dind-golang/test.yaml
```diff
@@ -0,0 +1,11 @@
+---
+docker:
+  entrypoint: top
+tests:
+  root_user_test: False
```
mishah334, line 5: @danielhoherd  this flag is controlling the root user test for particular docker image.

Makefile
```diff
@@ -21,3 +25,15 @@ show-quay-pull-urls: ## Show Quay.io pull for all images in repo
...
+.PHONY: build
+build:
+	docker-compose build ap-$(image_name)
+
+.PHONY: test
+test: export ASTRO_IMAGE_NAME = ap-$(image_name)
+test: export ASTRO_IMAGE_TAG = $(image_tag)
+test: export ASTRO_IMAGE_TEST_CONFIG_PATH = $(image_test_config)
+
+test:
```
danielhoherd, line 37: Need to add help text here so that `make help` works for this target.

Makefile
```diff
@@ -1,5 +1,9 @@
 .DEFAULT_GOAL := help
 
+image_name:=alertmanager
```
danielhoherd, line 3: Seems like we should delete this. Is it needed as a default for something?

bin/test.py
```diff
@@ -0,0 +1,109 @@
...
+# Read the test config
+if os.path.exists(ASTRO_IMAGE_TEST_CONFIG_PATH):
+    with open(ASTRO_IMAGE_TEST_CONFIG_PATH) as file:
+        config = yaml.safe_load(file)
+
+        if config is not None:
+
+            # Reading test config
+            if "tests" in config:
+                test_config = config["tests"]
```
danielhoherd, line 23: Change suggested by sourcery:
```suggestion
        if config is not None and "tests" in config:
            test_config = config["tests"]
```

bin/test.py
```diff
@@ -0,0 +1,109 @@
...
+    subprocess.check_call(["docker", "rm", "-f", docker_id])
+
+
+#################################################
+#                   User Test                   #
+#################################################
+
+
+@pytest.mark.skipif(
+    "root_user_test" not in test_config or False == test_config["root_user_test"],
```
danielhoherd, line 45:
```suggestion
    "root_user_test" not in test_config or test_config["root_user_test"] == False,
```

bin/test.py
```diff
@@ -0,0 +1,109 @@
...
+    # return a testinfra connection to the container
+    yield testinfra.get_host("docker://" + docker_id)
+    # cleanup container after test completion
+    subprocess.check_call(["docker", "rm", "-f", docker_id])
+
+
+#################################################
+#                   User Test                   #
+#################################################
+
```
danielhoherd, line 42: Rather than using large comment blocks like this, if we want to group tests into logical sections we should use classes or separate files.

postgresql/Dockerfile
```diff
@@ -30,5 +30,6 @@ USER root
 RUN addgroup postgres --gid 1001 && useradd -s /sbin/nologin --gid 1001 --uid 1001 postgres
 
 RUN chown -R 1001:1001 /opt/bitnami
+RUN chown -R 1001:1001 /bitnami/postgresql
```
danielhoherd, line 33: @pgvishnuram what do you think about this change? Seems like it has the potential to break existing installations.

requirements.txt
```diff
@@ -0,0 +1,3 @@
+pytest==7.1.1
+pytest-testinfra==6.0.0
+PyYAML==6.0
```
danielhoherd, line 3: We should rework this with [the requirements.in and `pip-compile` process we use in astronomer/astronomer](https://github.com/astronomer/astronomer/blob/f8eb4bbab437a7/Makefile#L60-L63) so the upgrade process is easier and more reliable.

.circleci/config.yml.j2
```diff
@@ -2,6 +2,8 @@
 {%- set docker_version = '20.10.11' -%}
 ---
 version: 2.1
+orbs:
+  python: circleci/python@2.0.3
```
danielhoherd, line 6: Do we really need this orb? Seems like it's a more complicated way to do `run: pip install requirements.txt`

danielhoherd (changes requested):
Looks great! I left some feedback that we should resolve before merging.

bin/test.py
mishah334: Yes, we can as only one condition for now to test here.

mishah334: Done

bin/test.py
mishah334: Done

Makefile
mishah334: I have removed the default value for them. Should we remove the declaration also?
`
image_name:=
image_tag:=latest
image_test_config:=$(image_name)/test.yaml
`

postgresql/Dockerfile
mishah334: Removed this permission. It was added earlier to provision permissions that allows to run docker image for debugging and testing. I worked to run them other way so no need for this permission further.

Makefile
mishah334: Done

bin/test.py
mishah334: Agree! Just removed them. I don't see much need to have class here.

requirements.txt
mishah334: Updated to use `pip-tools` as suggested.

.circleci/config.yml.j2
mishah334: Using Orb has also other benefit especially in case of this kind of build where we have multiple time same execution. Orb allow us to make reusable snippet. 

In our case we have to install pip dependencies to all test phase.

- Without orb, It will re-execute the pip install each test phase.
     - https://app.circleci.com/pipelines/github/astronomer/ap-vendor/922/workflows/5cc002cf-8910-4945-bdc0-85d527ba4fe2/jobs/34146 
     - https://app.circleci.com/pipelines/github/astronomer/ap-vendor/922/workflows/5cc002cf-8910-4945-bdc0-85d527ba4fe2/jobs/34149
- With orb, it will execute the pip install first with docker image test and then it will re-use it to all other.
    - https://app.circleci.com/pipelines/github/astronomer/ap-vendor/920/workflows/d781b5fa-97fa-4852-a03b-541abc7304f6/jobs/34132
    -  https://app.circleci.com/pipelines/github/astronomer/ap-vendor/920/workflows/d781b5fa-97fa-4852-a03b-541abc7304f6/jobs/34135

Here we don't have more package to install for now, so will be not much difference into execution time, but still it is suggested to use.

Makefile
danielhoherd: TIL you can do `make target var=val` and not just `var=val make target`

.circleci/config.yml.j2
danielhoherd: Man, they should update [their docs](https://circleci.com/developer/orbs/orb/circleci/python). They don't list any of those details.

danielhoherd (approved):

PR #287, opened 2022-04-26
Bump version to get a new base image build

danielhoherd:
## Description

- Bump fluentd "pre-release" version component so we get a rebuild of the existing fluentd version with a new base image

## Issues

https://github.com/astronomer/issues/issues/4521


vishwas-astro (commented):
Looks good

PR #304, opened 2022-05-23
vector 0.21.2-pre1

danielhoherd:
vector 0.21.2-pre1


shmanu017 (approved):

danielhoherd:
I'm just going to merge this even though the base image is reporting security vulnerabilities. The build will fail at first, but whenever the base image is updated then the build will succeed and the version will be released.

PR #312, opened 2022-05-27
Add new image of pgbouncer with custom gss support 

andriisoldatenko:
## Description

In this PR, I added new ap-pgbouncer-krb image

## Issues

- Resolves [#4700 ](https://github.com/astronomer/issues/issues/4700)


pgbouncer-krb/Dockerfile
```diff
@@ -14,21 +14,25 @@
 # limitations under the License.
 
 # https://github.com/CenterForOpenScience/docker-library/blob/master/pgbouncer/Dockerfile
-# https://quay.io/repository/centerforopenscience/pgbouncer
 FROM quay.io/astronomer/ap-base:3.15.4
```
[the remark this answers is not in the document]
andriisoldatenko: we can for sure. Will do.

pgbouncer-krb/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+      -e "s|logfile = |#logfile = |" \
+      -e "s|pidfile = |#pidfile = |" \
+      -e "s|listen_addr = .*|listen_addr = 0.0.0.0|" \
+      -e "s|auth_type = .*|auth_type = md5|" \
+      /etc/pgbouncer/pgbouncer.ini \
+    && cd .. \
+    && rm -rf /pgbouncer* \
+    && apk del .build-deps
+
+# COPY pgbouncer.ini users.txt /etc/pgbouncer
```
andriisoldatenko, line 73: TODO: remove this line

pgbouncer-krb/Dockerfile
```diff
@@ -0,0 +1,82 @@
...
+FROM quay.io/astronomer/ap-base:3.16.1
+LABEL maintainer="Astronomer <humans@astronomer.io>"
+
+ARG BUILD_NUMBER=-1
+LABEL io.astronomer.docker=true
+LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
+
+ENV REALM ${REALM:-EXAMPLE.COM}
+ENV KADMIN_PRINCIPAL ${KADMIN_PRINCIPAL:-kadmin/admin}
+ENV KADMIN_PASSWORD ${KADMIN_PASSWORD:-MITiys4K5}
```
andriisoldatenko, line 25:
```suggestion
ENV KADMIN_PASSWORD ${KADMIN_PASSWORD:-postgres}
```

pgbouncer-krb/entrypoint.sh
```diff
@@ -1,5 +1,6 @@
 #!/bin/sh
-# Based on https://raw.githubusercontent.com/brainsam/pgbouncer/master/entrypoint.sh
+# Based on https://github.com/edoburu/docker-pgbouncer/blob/master/entrypoint.sh
```
danielhoherd, line 2:
```suggestion
# Based on https://github.com/edoburu/docker-pgbouncer/blob/860ed3643/entrypoint.sh
```

pgbouncer-krb/entrypoint.sh
```diff
@@ -60,9 +61,10 @@ fi
 if [ ! -f ${PG_CONFIG_DIR}/pgbouncer.ini ]; then
   echo "Create pgbouncer config in ${PG_CONFIG_DIR}"
 
-# Config file is in �ini� format. Section names are between �[� and �]�.
-# Lines starting with �;� or �#� are taken as comments and ignored.
-# The characters �;� and �#� are not recognized when they appear later in the line.
+# Config file is in �ini� format. Section names are between �[� and �]�.
```
danielhoherd: What's up with these `¿ini¿` things? Shouldn't that be `"ini"` or something?

pgbouncer-krb/Dockerfile
andriisoldatenko: done.

pgbouncer-krb/entrypoint.sh
andriisoldatenko: @danielhoherd fixed.

pgbouncer-krb/entrypoint.sh
```diff
@@ -0,0 +1,205 @@
...
+fi
+
+if [ ! -f ${PG_CONFIG_DIR}/pgbouncer.ini ]; then
+  echo "Create pgbouncer config in ${PG_CONFIG_DIR}"
+
+# Config file is in "ini" format. Section names are between "[" and "]".
+# Lines starting with ";" or "#" are taken as comments and ignored.
+# The characters ";" and "#" are not recognized when they appear later in the line.
+# shellcheck disable=SC2059
+  printf "\
```
danielhoherd, line 115: This would be easier to maintain as a HEREDOC:
```suggestion
cat <"EOF" > "${PG_CONFIG_DIR}/pgbouncer.ini"
```
then strip out the line endings and just do it exactly like you'd see in the config with shell variables interpolated.

andriisoldatenko: fixed in 686abd4

danielhoherd (approved):
LGTM. that `${FOO:+foo = ${FOO}}` syntax is really sweet! I had never seen that before.

pgbouncer-krb/entrypoint.sh
danielhoherd: Oh, I think I see now why they did it with `printf`. Using `printf` (or `echo -en`), you have to specify the line ending, which lets you skip the line ending if the var is not printed. That might be worth undoing. The way it is now, the config file will be full of empty lines for any vars that are not defined.

danielhoherd: Yeah, heredoc has the fewest empty lines:

```
#!/usr/bin/env bash

echo "=========== cat syntax"
cat <<EOF
${FOO:+foo = ${FOO}}
${SHELL:+shell = ${SHELL}}
${BAZ:+baz = ${BAZ}}
${HOME:+home = ${HOME}}
EOF

echo "=========== echo syntax"
echo -en "
${FOO:+foo = ${FOO}\n}
${SHELL:+shell = ${SHELL}\n}
${BAZ:+baz = ${BAZ}\n}
${HOME:+home = ${HOME}\n}"

echo "=========== printf syntax"
printf "
${FOO:+foo = ${FOO}\n}
${SHELL:+shell = ${SHELL}\n}
${BAZ:+baz = ${BAZ}\n}
${HOME:+home = ${HOME}\n}"
$ ./test.sh
=========== cat syntax

shell = /opt/homebrew/bin/bash

home = /Users/danielh
=========== echo syntax


shell = /opt/homebrew/bin/bash


home = /Users/danielh
=========== printf syntax


shell = /opt/homebrew/bin/bash


home = /Users/danielh
```

andriisoldatenko: yeah it's okey. Big win when using `cat <<EOF` i removed ignore shellcheck `disable=SC2059`.
 So now we enable all checks. Which is good.

PR #320, opened 2022-06-03
Unique grafana uids, grafana 8.5.4

danielhoherd:
While investigating a grafana crash loop in prod, I found an error about duplciate uids in dashboard:

```
logger=provisioning.dashboard type=file name=default t=2022-06-03T23:07:55.76+0000 lvl=eror msg="failed to save dashboard" error="A dashboard with the same uid already exists"
```

Upgrading to the latest version of our grafana image did not solve it. I discovered that I had added a new version of a dashboard that had a duplicate UID. This PR fixes that, and upgrades to grafana 8.5.4


shmanu017 (commented):
LGTM

PR #325, opened 2022-06-16
Vector 0.22.2. Use curl for downloader image.

danielhoherd:
- Vector 0.22.2
- Use curl for downloader image because it's only 11mb vs 892mb for python:3.10-buster


pgvishnuram (approved):

danielhoherd:
We should hold off on merging this until CI passes, whenever there's a new build of python:3.10-slim

rishkarajgi (approved):

rishkarajgi:
> We should hold off on merging this until CI passes, whenever there's a new build of python:3.10-slim

2 critical issues found

danielhoherd:
I have an alert set up so we'll know when a new base image is released.

PR #386, opened 2022-08-03
update postgresql base image

pgvishnuram:
## Details

Updating postgresql 11.15 base image with debian-10-r93

## Related Issues
https://github.com/astronomer/issues/issues/4881

## Checklist

- [x] version.txt was updated


postgresql/Dockerfile
```diff
@@ -15,7 +15,7 @@
 
 # https://hub.docker.com/r/bitnami/postgresql/tags/
 
-FROM bitnami/postgresql:14.4.0-debian-11-r7
+FROM bitnami/postgresql:11.15.0-debian-10-r93
```
vishwas-astro, line 18: @pgvishnuram We are going from 14.4 to 11.15 only because everything is released from main and there are no release branches per version?

pgvishnuram: Yes vishwas - we usually make changes and update version to get pushed to quay

PR #422, opened 2022-10-03
Add executor resource profiling to stats mappings

jarfgit:
## Details

The astro runtime executor emits stats. This PR adds corresponding mappings to receive them.

## Related Issues
https://github.com/astronomer/stellar/issues/292

## Checklist

- [x] version.txt was updated


statsd-exporter/include/mappings-gen2.yml
```diff
@@ -101,6 +101,14 @@ mappings:
     name: "airflow_executor_queued_tasks"
     labels: {}
 
+  - match: airflow.executor.*.resources.*
```
conorbev, line 104: I think that this line is matching [this metric](https://github.com/astronomer/stellar/blob/59e49dc5030ebc0e50240a9a1e34edde7ee44379/stellar/allocator/allocator.py#L493) ?

If so I don't think it would be prefixed with `airflow.` ? Wouldn't this be: `- match: executor.*.resources.*`

statsd-exporter/include/mappings-gen2.yml
```diff
@@ -101,6 +101,14 @@ mappings:
...
+    name: airflow_runner_resources
+    labels: {}
```
conorbev, line 106: If labels are left empty like this then there will just be a single metric in chronosphere named `airflow_runner_resources` with a mix of different `target_name`s and `resource`s.

Shouldn't we have labels like:

```
target_name: "$1"
resource: "$2"
```

to extract those labels from the named labels?

statsd-exporter/include/mappings-gen2.yml
jarfgit: I have a corresponding PR on stellar to update these metrics and ensure they are unique to the mapping.yml. https://github.com/astronomer/stellar/pull/303

jarfgit: I have a corresponding PR to update these names on the executor.

statsd-exporter/include/mappings-gen2.yml
jarfgit: Ah, ok. That's what the labels are more! Makes sense. Will push an update.

danielhoherd:
@jarfgit because this is gen2 can you tag some ops folks from gen2 to approve this? @jpweber and @goutamtadi1 may be good to tag for review. This doesn't affect anything in gen1, which is what the ap-vendor maintainers work on.

pgvishnuram:
@jarfgit - pls update version.txt as well

pgvishnuram (approved):

goutamtadi1 (approved):

goutamtadi1:
> @goutamtadi1 I should call out that stellar is able to emit both statsd and dogstatsd metrics depending on how we set `AIRFLOW__METRICS__STATSD_DATADOG_ENABLED`. Does this mean we need to add mappings to https://github.com/astronomer/ap-vendor/blob/main/dogstatsd/include/dogstatsd.yaml?

I don't think so. we only forward metrics from statsd to datadog unless until datadog is default and only metrics exporter.
But @howardyoo might have some ideas

howardyoo:
> > @goutamtadi1 I should call out that stellar is able to emit both statsd and dogstatsd metrics depending on how we set `AIRFLOW__METRICS__STATSD_DATADOG_ENABLED`. Does this mean we need to add mappings to https://github.com/astronomer/ap-vendor/blob/main/dogstatsd/include/dogstatsd.yaml?
> 
> I don't think so. we only forward metrics from statsd to datadog unless until datadog is default and only metrics exporter. But @howardyoo might have some ideas

No, we do not need to add these mappings to dogstatsd.yaml file.

jarfgit:
> I don't think so. we only forward metrics from statsd to datadog unless until datadog is default and only metrics exporter.

@goutamtadi1 I need clarification here. Why are we forwarding some metrics to ddog? If stellar is emitting metrics in a datadog format, will they be mapped correctly by the statsd_exporter?

The broader context for my question is that we are trying to wire up metrics from stellar in dogstatsd format. I suspect there may be more to this than merging this PR. If so, I want to understand exactly what we need to do and set timeline / expectations accordingly.

cc @howardyoo, @da-w-xu @conorbev

PR #427, opened 2022-10-11
Improving Python Script to Build & Push Docker Image

mishah334:
## Details

Currently, Docker Image tags are overwritten when they should not be. This happens because the current script is failing on check because of Authentication. This was previously fixed in PR. 

The changes here are about adding more features.

- Adding capability to add `overwrite_tags` capability from CircleCI.
- Adding CircleCI Job to validate tag for non `main` branch.
- Improved the script to support multiple tags (comma-separated).

**Previous PRs:**
- https://github.com/astronomer/ap-vendor/pull/425
- https://github.com/astronomer/ap-vendor/pull/426

## Related Issues

Issue: https://github.com/astronomer/issues/issues/5084

## Checklist

- [ ] version.txt was updated

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,7 +15,7 @@ parameters:
   overwrite_tags:
     type: boolean
     default: false
-    description: "Set True if you want to override existing docker image tag."
+    description: "Set True if you want to overwrite existing docker image tag."
 
 workflows:
   version: 2.1
--- a/.circleci/config.yml.j2
+++ b/.circleci/config.yml.j2
@@ -15,7 +15,7 @@ parameters:
   overwrite_tags:
     type: boolean
     default: false
-    description: "Set True if you want to override existing docker image tag."
+    description: "Set True if you want to overwrite existing docker image tag."
 
 workflows:
   version: 2.1


bin/docker-operations.py
```diff
@@ -22,32 +22,84 @@
...
+    repository: str,
+    image: str,
+    tags: list,
+    override_tags: bool,
+):
+    docker_image_uri = registry + "/" + repository + "/" + image
+
+    if override_tags:
+        print(
+            "INFO: Override is set to True, if the tag will already present it will get override."
```
danielhoherd, line 66:
```suggestion
            "INFO: Overwrite is set to True. If the tag already exists it will be overwritten."
```

bin/docker-operations.py
```diff
@@ -22,32 +22,84 @@
...
+                    docker_client.images.get_registry_data(
+                        name=(docker_image_uri + ":" + tag)
+                    )
+
+                    print(
+                        f"INFO: The docker tag {docker_image_uri}:{tag} already exists. Skipping the Docker push!"
+                    )
+                except APIError as dokerAPIError:
+                    print(
+                        f"INFO: Image tag {docker_image_uri}:{tag} not found on server. It will add to push list."
```
danielhoherd, line 88: Consistent wording with the prior INFO message
```suggestion
                        f"INFO: Docker tag {docker_image_uri}:{tag} not found on server. It will be added to the push list."
```

danielhoherd (approved):
Looks great!

PR #467, opened 2022-12-22
Grafana

shmanu017:
## Details

- Fix Grafana shows unhealthy airflow state due to kube jobs
- Update Grafana subversion

## Related Issues

https://github.com/astronomer/issues/issues/2208

## Checklist

- [X] version.txt was updated


grafana/include/airflow-state.json
```diff
@@ -1784,6 +1784,6 @@
   },
   "timezone": "",
   "title": "Airflow State",
-  "uid": "h7dUiVJmz",
+  "uid": "h7dUiVJmq",
```
danielhoherd, line 1787: We should keep this the same. If we change it, it can break people's bookmarks. EG: https://grafana.dev.astrodev.io/d/h7dUiVJmz/airflow-state

danielhoherd (approved):

PR #473, opened 2023-01-19
Add k6 image for static fire

rujhan-arora-astronomer:
## Details

Add k6 image for static fire

## Related Issues

5354

## Checklist

- [x] version.txt was updated


pgvishnuram:
@danielhoherd - should this k6 load test image can we add it here

static-fire/trivyignore
```diff
@@ -0,0 +1,3 @@
+# Upstream registry service include several CVEs.
+# We should try to remove these each time we update.
+CVE-2022-32149
```
danielhoherd, line 3: Does this apply to this static-fire? This looks like it was copied from registry but is not needed here. If it's needed we should improve this comment text and make it more accurate for the static-fire project.

rujhan-arora-astronomer: Yeah, I'll remove it

rujhan-arora-astronomer:
> @rujhan-arora-astronomer i agree with what @pgvishnuram is trying to say: this docker image doesn't really belong in ap-vendor. i think the docker build and push should happen from the static-fire repo itself. we can help you set that up if you need.

This is just a generic image, with Node.js + K6 installed. Fundamentally, not coupled with static fire repo. Also, we don't need to build this image again. This is a one time thing (or atleast, the changes will be rare)

danielhoherd:
> > @rujhan-arora-astronomer i agree with what @pgvishnuram is trying to say: this docker image doesn't really belong in ap-vendor. i think the docker build and push should happen from the static-fire repo itself. we can help you set that up if you need.
> 
> This is just a generic image, with Node.js + K6 installed. Fundamentally, not coupled with static fire repo. Also, we don't need to build this image again. This is a one time thing (or atleast, the changes will be rare). I raised this PR, while on call with @pgvishnuram

Images in the repo are shipped to customers and are maintained for security updates. The image you want to create does not suit that profile. The quay repo is quay.io/astronomer/static-fire, so it makes sense that it would ship from that repo. If it's a general purpose image, we can delete quay.io/astronomer/static-fire and make quay.io/astronomer/k6-node or something, but I doubt we want to do that.

rujhan-arora-astronomer:
> > > @rujhan-arora-astronomer i agree with what @pgvishnuram is trying to say: this docker image doesn't really belong in ap-vendor. i think the docker build and push should happen from the static-fire repo itself. we can help you set that up if you need.
> > 
> > 
> > This is just a generic image, with Node.js + K6 installed. Fundamentally, not coupled with static fire repo. Also, we don't need to build this image again. This is a one time thing (or atleast, the changes will be rare). I raised this PR, while on call with @pgvishnuram
> 
> Images in the repo are shipped to customers and are maintained for security updates. The image you want to create does not suit that profile. The quay repo is quay.io/astronomer/static-fire, so it makes sense that the docker image would ship from that github repo. If it's a general purpose image, we can delete quay.io/astronomer/static-fire and make quay.io/astronomer/k6-node or something, but I doubt we want to do that.

Yeah, we can remove it if it doesn't fit in. Though quay.io/astronomer/static-fire seems like an image of the static-fire repo itself, which we don't want. We just want to put this k6+node image in quay and use that in static fire workflow build. What would be the appropriate way to do that?

danielhoherd:
I worked with Rujhan to add CircleCI build and push to the static-fire repo.

PR #475, opened 2023-01-20
Use looser versions so we inherit security fixes

danielhoherd:
## Details

The vector 0.26 build failed because of a python CVE that was fixed in a newer version of the python docker image. Since we don't really care about the specific version of python or curl, I have loosened these versions up so we can automatically get security fixes. The thing here that we really care about the version of is vector, which is pinned exactly.

## Related Issues

N/A

## Checklist

- [ ] version.txt was updated (not needed, previous build failed)


danielhoherd:
Actually we need this open to ship 0.26.0.

PR #485, opened 2023-02-03
Test for pod securityContexts running as non-root 

shmanu017:
## Details

 Pod securityContexts running as non-root. We need to specify a UID instead of a USER in order for k8s to ensure the user is not running as UID 0 (root).

## Related Issues

https://github.com/astronomer/issues/issues/5339

## Checklist

- [X] version.txt was updated


danielhoherd (commented):
LGTM

danielhoherd (approved):
LGTM.

PR #497, opened 2023-03-03
Fix airflow state dashboard health by excluding airflow-cleanup-pods

danielhoherd:
## Details

Fix the airflow-state dashboard by excluding airflow-cleanup-pods pods. I've excluded them entirely because if I only exclude the ones that have completed, it looks as if the cleanup pod is still running, which is not accurate.

Use grafana 8.5.21

## Related Issues

https://github.com/astronomer/issues/issues/4449

## Testing

I have manually tested this in prod in the query edit view to verify that these changes function as expected.

## Checklist

- [x] version.txt was updated


shmanu017 (approved):

danielhoherd:
There are CVEs in grafana 8.5.21, so we should wait until there is a newer version and update to it before merging this. <https://hub.docker.com/r/grafana/grafana/tags?page=1&name=8.>

grafana/Dockerfile
```diff
@@ -14,7 +14,7 @@
 # limitations under the License.
 
 # https://hub.docker.com/r/grafana/grafana/tags
-FROM grafana/grafana:8.5.16
+FROM grafana/grafana:8.5.21
```
pgvishnuram, line 17: 8.5.21 - seems not publically showing in grafana releases

danielhoherd: https://hub.docker.com/layers/grafana/grafana/8.5.21/images/sha256-5419f6affd2c34fce914bd1787985c0a27435d13c39dd535ba2121702e0609cb?context=explore

danielhoherd:
I merged these changes to https://github.com/astronomer/ap-vendor/pull/506 so I'm closing this.

Issue #507, opened 2023-03-31
remap metric airflow.local_task_job.* from incoming metrics list


howardyoo:
# Problem
When datadog integration is enabled for Astro, you may observe the following metrics in the metrics summary list of datadog:
<img width="517" alt="image" src="https://user-images.githubusercontent.com/32691630/229220002-b97dbd7a-c5f4-45b5-9f82-85056163d115.png">
The metric `airflow.local_task_job.*` is mentioned in airflow documentation as such: https://airflow.apache.org/docs/apache-airflow/2.5.1/administration-and-deployment/logging-monitoring/metrics.html#counters, however, there is little use for it, and moreover, this metric is not part of the metric list that datadog airflow integration has in its documentation (https://docs.datadoghq.com/integrations/airflow/?tab=host#metrics).

Also, this metric is not something that is frequently used, since the metric is only emitted when a task is explicitly run via airflow task run command, and also, the metric is only used to check whether the metric has run or got stuck somewhere (if it is stuck, you won't see the metric itself, and if it exited with some exit code, you should see the exit code other than zero).

This metric is also problematic as it introduces high cardinality metrics - and also occupies metric namespace heavily.

# Solution
Add a mapping rule in dogstat's default rule config - such that it will remove these metrics from being emitted.
```yaml
      - match: 'airflow.local_task_job.task_exit.*.*.*.*'
        name: "airflow.local_task_job.task_exit"
        tags:
          dag_id: "$2"
          task_id: "$3"
          exit_code: "$4"
```

howardyoo:
Looks like the mapping rule in dogstatsd (or datadog agent) does not have means to completely `block` certain metrics from coming in, which is kind of silly (like every agent should have that kind of feature). Before we have a better means to control this, I'd say the secondary option would be to reduce the cardinality of this metrics.

PR #509, opened 2023-04-03
Additional mapping rule to dogstatsd to control cardinality of local_task_job metric

howardyoo:
This PR adds an entry to ap-vendor's dogstatsd image by applying additional rule to re-map the metric mentioned in #507 

## Details

The changes are defined in #507, but to briefly explain, we're adding the following entry to the dogstatsd.yaml
```yaml
      - match: 'airflow.local_task_job.task_exit.*.*.*.*'
        name: "airflow.local_task_job.task_exit"
        tags:
          dag_id: "$2"
          task_id: "$3"
          exit_code: "$4"
```
This will reduce the multiple metric names shown up in the datadog metrics list
image
into a single metric called `airflow.local_task_job.task_exit` with dag_id, task_id, and exit_code as tags, and removing the job_id part.

## Related Issues

#507 

## Checklist

- [x] version.txt was updated


danielhoherd:
@howardyoo you should add some reviewers that work with dogstatsd. @pgvishnuram and I do not use dogstatsd, as it is not a gen-1 component, which is the product we work on.

howardyoo:
> @howardyoo you should add some reviewers that work with dogstatsd. @pgvishnuram and I do not use dogstatsd, as it is not a gen-1 component, which is the product we work on.

I don't know of anyone who works with the dogstatsd, except myself - and I did test everything, so it should all be good?

goutamtadi1 (approved):
LGTM. Were you able to verify the behavior in real time?

howardyoo:
yes, I basically ran a docker image of dogstatsd in my local, applied the
updated yaml, and submitted the same statsd data to verify it.

PR #510, opened 2023-04-07
Make 'release' require 'validate-tags'

danielhoherd:
## Details

Make 'release' require 'validate-tags'.

The current flow looks like this:

Screenshot 2023-04-07 at 6 28 30 PM

After merging these changes, "validate-tags" must pass before "release" can run.

## Related Issues

N/A

### DIFF
--- a/.circleci/continue-config.yml
+++ b/.circleci/continue-config.yml
@@ -202,18 +202,6 @@ workflows:
           directory: alertmanager
           requires:
             - test
-      - release:
-          directory: alertmanager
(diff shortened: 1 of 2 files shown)


shmanu017 (approved):

danielhoherd:
This was a mistake. I didn't understand the whole flow for all branches. I am going to revert this. The main point of confusion is that we do not usually release from feature branches, but in my case I was, which caused the steps to appear like they had incorrect relationships.

PR #511, opened 2023-04-11
rework vector to handle heartbeat from primary container

danielhoherd:
## Details

This branch reworks vector to deal with an edge case where the primary container is brutally killed. This would leave the vector logging sidecar in a zombie state forver. The new implementation watches for a heartbeat in a file in the shared volume. The "heartbeat" file will show the last timestamp that the airflow container emitted into the heartbeat file, which elsewhere is implemented to be every 5 seconds.

Since we are now signaling from airflow using the shared log volume, I removed the HTTP server and set up a mechanism for airflow to signal to the sidecar that it has exited by creating a "finished" file.

Here are some cases that are accounted for in the code:
- If the "heartbeat" file doesn't exist, this script waits for it to show up.
- If the "heartbeat" file never shows up, this script will only exit when the "finished" file shows up. (this could be seen as a design flaw)
- If the "heartbeat" file disappears or becomes empty, the last known heartbeat is used.
- If the "finished" file is seen, the script will ask vector to quit and then exit.
- If the script is killed, vector is terminated.

The version of vector we were using, 0.26, was a bit old. I bumped it to 0.28.2 which is newer but not the newest (0.29.0 came out yesterday), and will give us a quick identifier of whether or not the image uses filesystem signaling or HTTP signaling. I tested this version to make sure it's working with our default configs, and I looked through the release notes for breaking changes and didn't find anything that looked like it would affect us.

I also added `bin/validate-semver.py` which helped me during debugging and releasing test versions. This script could use some improvement to also validate that the tags are docker compatible, but it's better than it was before.

This change goes hand in hand with the vector changes in this PR https://github.com/astronomer/astronomer/pull/1855

## Related Issues

- https://github.com/astronomer/issues/issues/5469
- https://github.com/astronomer/astronomer/pull/1855


.circleci/continue-config.yml.j2
```diff
@@ -54,7 +54,9 @@ workflows:
             - scan-trivy
           filters:
             branches:
-              only: main
+              only:
+                - main
+                - 5469-rework-vector
```
danielhoherd, line 59: Before merging we need to delete this:
```suggestion
```

vector/version.txt
```diff
@@ -1 +1 @@
-0.26.0-4
+0.0.5469-4
```
danielhoherd, line 1: Before merging we need to fix this.
```suggestion
0.26.0-5
```

vector/sidecar_server.py
```diff
@@ -1,43 +1,72 @@
...
-        if errs:
-            print(f"{errs=}")
+        quit_proc()
         raise SystemExit()
 
 
-class ThisServer(server.HTTPServer):
+class MessageServer(server.HTTPServer):
+    heartbeat_file = Path("/var/log/sidecar-log-consumer/heartbeat")
+    heartbeat_max_age = 180  # seconds
```
danielhoherd, line 45: I'm not sure this is the right number. Maybe we could reduce it to 120?

vector/sidecar_server.py
```diff
@@ -1,43 +1,72 @@
...
+            # Sometimes the file contents are empty due to a race condition, so we only update
+            # last_timestamp if the file contents can be converted to a float.
+            if last_timestamp := float(self.heartbeat_file.read_text()):
+                self.last_timestamp = last_timestamp
+            age = time.time() - self.last_timestamp
+            if age > self.heartbeat_max_age / 2:
+                print(
+                    f"WARNING: Heartbeat has not been sent for {age:0.1f} seconds",
+                    flush=True,
+                )
```
danielhoherd, line 63: This section is really chatty. It produces one message every 0.5 seconds. I'm not sure it's even needed.

rishkarajgi (approved):

PR #525, opened 2023-04-24
update USER with uid for openresty service

pgvishnuram:
## Details

Do not merge this PR until this description is replaced with relevant info about what images are being updated and what changes are being made.

## Related Issues

## Checklist

- [ ] version.txt was updated


openresty/Dockerfile
```diff
@@ -36,4 +36,4 @@ RUN mkdir -p /usr/local/openresty/nginx/client_body_temp \
 # upgrade apk packages
 RUN apk upgrade --no-cache
 
-USER nobody
+USER 65534
```
rishkarajgi, line 39: what does this number correspond to? Is it a random static id we are assigning?

pgvishnuram: This is default nobody user available in docker image

```nobody:x:65534:65534:nobody:/:/sbin/nologin``` - since our securityContext is set to runAsNonRoot: true it now needs uid not username

rishkarajgi (approved):

PR #528, opened 2023-05-06
Update mappings-gen2.yml

conorbev:
There are new Airflow metrics around Triggerers and Dataset usage that would give us enhanced visibility into user adoption of those Airflow features. This PR updates our statsd config to allow those new metrics to flow to Chronosphere.

The new Dataset metrics were requested by the Astronomer Data team here: https://astronomer.slack.com/archives/C02PABPU6B0/p1673028075337969

The new Triggerer metrics are discussed here: https://astronomer.slack.com/archives/CGQSYG25V/p1683233844842789

## Details

This change modifies the configuration of our [Prometheus StatsD Exporter](https://github.com/prometheus/statsd_exporter) which defines which Airflow metrics will be scraped by Chronocollector. 

## Checklist

- [ ] version.txt was updated <-- NB I am not familiar with version.txt, could I please get help from a reviewer if this is needed for this change?


conorbev:
Thanks @danielhoherd I incremented the version. Is that OK? Perhaps you meant it should have been `0.23.1-1` ?

@jpweber Would you mind giving this a quick look too?

statsd-exporter/version.txt
```diff
@@ -1 +1 @@
-0.23.1
+0.23.2
```
danielhoherd, line 1: We're rebuilding 0.23.1 https://github.com/astronomer/ap-vendor/blob/74cfcf4a1e3dbb989245f936dbf7808e62645c0b/statsd-exporter/Dockerfile#L17 so this should be:

```suggestion
0.23.1-1
```

danielhoherd (approved):

pgvishnuram:
@danielhoherd - this got merged with pre-commit failures is this expected

PR #536, opened 2023-05-23
Merge requirements files and use a venv for local dev

danielhoherd:
## Details

While implementing tests in another PR, we ran into some inconsistencies with local build and CI build. To reduce the variety between those two environments, I've merged the requirements files and install them into a venv, then use that venv for all operations in the Makefile that need to execute python scripts within this repo.

To bumped the redis version so we can see a build pipeline succeed before merging these changes.

## Related Issues

N/A


redis/version.txt
```diff
@@ -1 +1 @@
-6.2.7-1
+6.2.7-2
```
ashb, line 1: Did you mean to change this version here?

ashb (approved):

redis/version.txt
ashb: Ah to test things, gotcha

PR #537, opened 2023-05-24
Switch to using `python-on-whales` (a docker CLI wrapper) for building

ashb:
## Details

The official docker python client doesn't support building via Buildx,
which means that there are differences between `make build` and how CI
was building that.

This library doesn't have full support for all the operations we want
(for example it can't look at tags on a remote registry) so it has only
been swapped in for the `build` subcommand

To remove future differences and sources of confusion the Makefile rule
has been changed to build the same way that CI does.

## Related Issues

Closes astronomer/issues#5639

## Checklist

- [ ] Will rebase my ap-postgresql change on top of this to test


ashb:
Looks like the fail is just because in this PR I haven't updated any version:


image

PR #574, opened 2023-07-26
Dev updates

danielhoherd:
## Details

Bump a bunch of dev components. This is in response to a dependabot PR to patch a CVE https://github.com/astronomer/ap-vendor/pull/572

Bump the version in redis so we can see a build pipeline succeed with the new changes.

## Checklist

- [X] version.txt was updated


pgvishnuram:
also lets update k8s versions too

danielhoherd:
> also lets update k8s versions too

Can you be more specific? I don't see anything k8s related in here.

PR #578, opened 2023-08-18
Delete all datasources from grafana dashboards.

danielhoherd:
## Details

Delete all datasources from grafana dashboards. This was done with the following command pipeline:

```sh
git grep -l PBFA97CFB590B2093 |
while read -r F ; do
  gron "$F" |
  grep -vF '].datasource' |
  gron -u > asdf && \
  mv asdf "$F"
done
```

This may be too heavy handed. For example, it deleted blocks like this:

```
"datasource": {
  "type": "datasource",
  "uid": "grafana"
},
```

~I'd imagine these blocks are not needed, but I'm not 100% certain.~ I built this container and tested it in dev. I checked the "Airflow Resource Utilization" dashboard, which had removed the grafana datasource, and almost everything looked fine. One odd thing was that I got a permission error when accessing the `Astronomer Houston` and `Fluentd 2022-02` dashboards.

One interesting thing is that if we edit dashboards using the GUI and then export them, they still contain the problematic data source. We'll have to keep an eye out for that in the future.

## Related Issues

https://github.com/astronomer/issues/issues/5727

## Checklist

- [x] version.txt was updated


pgvishnuram:
can you also test this by deploying in one test cluster

danielhoherd:
I've filed a follow-up issue about the permissions problem. It is not tied to this change. https://github.com/astronomer/issues/issues/5759

pgvishnuram (approved):
lets merge this and share it to customer to unblock their issue

PR #587, opened 2023-09-12
update registry version 2.8.2 with alpine 3.18.3

pgvishnuram:
## Details

This PR syncs registry distribution service with upstream version( 2.8.2) and bring newer improvements and features.

## Related Issues
https://github.com/astronomer/issues/issues/5772

## Checklist

- [x] version.txt was updated


rishkarajgi (approved):

danielhoherd (approved):
This is a rather complicated build process with some components that are not at all obvious. I can't wait to get back on track with upstream, but I guess for now this will have to do.

PR #591, opened 2023-09-13
ap-init curl

danielhoherd:
## Details

Add curl to ap-init. Currently it only has busybox wget, which is extremely limited in features.

## Related Issues

N/A

## Checklist

- [x] version.txt was updated


rishkarajgi (approved):

rishkarajgi:
How does adding curl benefit here exactly?

pgvishnuram:
> How does adding curl benefit here exactly?

instead of using ap-kibana we will be using this image

danielhoherd:
@rishkarajgi 

> Currently it only has busybox wget, which is extremely limited in features.

busybox has:
```
/docker-share $ wget --help
BusyBox v1.36.1 (2023-07-27 17:12:24 UTC) multi-call binary.

Usage: wget [-cqS] [--spider] [-O FILE] [-o LOGFILE] [--header STR]
	[--post-data STR | --post-file FILE] [-Y on/off]
	[-P DIR] [-U AGENT] [-T SEC] URL...

Retrieve files via HTTP or FTP

	--spider	Only check URL existence: $? is 0 if exists
	--header STR	Add STR (of form 'header: value') to headers
	--post-data STR	Send STR using POST method
	--post-file FILE	Send FILE using POST method
	-c		Continue retrieval of aborted transfer
	-q		Quiet
	-P DIR		Save to DIR (default .)
	-S    		Show server response
	-T SEC		Network read timeout is SEC seconds
	-O FILE		Save to FILE ('-' for stdout)
	-o LOGFILE	Log messages to FILE
	-U STR		Use STR for User-Agent header
	-Y on/off	Use proxy
```
and nothing else. No `--retry`, `--dump-header`, `--head`, etc.. Curl will be useful in an init container.

PR #610, opened 2023-09-26
upgrade KEDA and KEDA api-server to 2.11.2

AmFlint:
## Details

In this PR, I upgraded the version of KEDA we want to use in Astronomer to version 2.11.2.

We want to mirror this image for multiple reasons:
- Latest version available at the time of creating this PR.
- Supports K8s 1.24 and above (all clusters in production are on 1.24 and above).
- This version fixes the few CRITICAL vulnerabilities we found on KEDA 2.9.3.

## Related Issues

https://github.com/astronomer/astro/issues/16478

## Checklist

- [x] version.txt was updated


danielhoherd (approved):
This looks fine. We do not use KEDA in "Software" so no action is needed there.

PR #663, opened 2023-12-26
add conditional config for update ca-certificates

pgvishnuram:
## Details

Adds conditional update ca-certificates logic to registry entrypoint to be invoked when global private ca feature is enabled

## Related Issues

https://github.com/astronomer/issues/issues/5939

## Checklist

- [x] version.txt was updated


registry/docker-entrypoint.sh
```diff
@@ -2,6 +2,14 @@
 
 set -e
 
+# In production, we mount the registry TLS certificate
+# into /usr/local/share/ca-certificates. This completes
+# the certificate installation so we can trust the registry.
+if [ "$UPDATE_CA_CERTS" = "true" ]; then
+echo "Running update-ca-certificates"
+update-ca-certificates
+fi
```
pgvishnuram, line 11: if [ "$UPDATE_CA_CERTS" ]; then => this also works @danielhoherd let me know on this

danielhoherd: Seems fine.

registry/Dockerfile
```diff
@@ -52,6 +52,7 @@ ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 
+ENV UPDATE_CA_CERTS false
```
danielhoherd, line 55: /bin/sh does not work this way. `if [ "$UPDATE_CA_CERTS" ]` will pass if the variable is defined, regardless of its contents.

pgvishnuram: then can i merge with current changes

danielhoherd: We should look at how other docker containers do this type of check. It's probably something more complicated than a simple 'if defined'

registry/docker-entrypoint.sh
danielhoherd: We need to change this logic before we merge this if we are looking at the value of the variable instead of only if it is defined.

registry/Dockerfile
pgvishnuram: this variable will be added in registry statefulset as well so when globalprivatecerts is enabled UPDATE_CA_CERTS=this is set to true so this setting will be overrided at all times

danielhoherd: What I mean though is if we are setting it to `false` that implies that setting it to `true` will change the behavior, but that's not the way the shell test you implemented works. If you are going to use a value like "true" or "false" then the logic should actually operate on those values.

If we want to use `true` and `false` we should do something like this:

```sh
#!/bin/sh

check_boolean_value() {
  if echo "$1" | grep -qiE "^(true|yes|1)$" ; then
    echo "$1 true"
    return 1
  else
    echo "$1 false"
    return 0
  fi
}


check_boolean_value 1
check_boolean_value 0
check_boolean_value true
check_boolean_value FALSE1
check_boolean_value yes
check_boolean_value no
check_boolean_value NO
check_boolean_value TrUE
check_boolean_value jksldf
check_boolean_value zero
check_boolean_value ''
```

which returns
```
1       true
0       false
true    true
FALSE1  false
yes     true
no      false
NO      false
TrUE    true
jksldf  false
zero    false
false
```

pgvishnuram: https://github.com/astronomer/houston-api/blob/main/bin/entrypoint#L19-L25

pgvishnuram: another reference https://github.com/astronomer/ap-vendor/blob/main/postgres-exporter/include/entrypoint#L10C1-L14

registry/docker-entrypoint.sh
pgvishnuram:
```sh
if [ "$UPDATE_CA_CERTS" = "true" ] || [ "$UPDATE_CA_CERTS" = "True" ] ; then
echo "Running update-ca-certificates"
fi
```

our purpose is to run this command when privateCA is enabled

registry/Dockerfile
pgvishnuram:
```sh
if [ "$UPDATE_CA_CERTS" = "true" ] || [ "$UPDATE_CA_CERTS" = "True" ] ; then
echo "Running update-ca-certificates"
fi
```
our purpose is to run this command when privateCA is enabled

danielhoherd (approved):
This code seems to make sense and work as described.

PR #664, opened 2024-01-12
Add optional config to keep high cardinality metrics

stuart23:
## Details

Some customers are relying on the metrics that they were receiving in prior to version 0.16.0

This PR adds an extra statsd-exporter config that was previously in production before [this PR](https://github.com/astronomer/ap-vendor/pull/37/files). Instructions on how to configure software to use this config can be found in the Knowledge Base.

This does not change the upstream version number, so I am not updating the version.txt

## Checklist

- [x] version.txt was updated


danielhoherd:
I fixed:
- pre-commit auto-formatting
- updated the twistcli ignores
- Bumped the version, which is required in order to publish a new image.

stuart23:
Let's just hold on this for a moment, because it looks like we can just spec the mappings in helm now https://github.com/astronomer/airflow/blob/main/chart/templates/configmaps/statsd-configmap.yaml#L43

statsd-exporter/include/mappings-high-cardinality.yml
```diff
@@ -0,0 +1,75 @@
+# This config undoes the changes done in https://github.com/astronomer/ap-vendor/pull/37/files
+# so that we can continue to support customers who need these metrics still.
+
```
danielhoherd, line 3: I think we should get rid of this text. This PR is suitable for giving us context about why this file exists.
```suggestion
```

danielhoherd:
@stuart23 do we still need this?

PR #665, opened 2024-01-16
update grafana version 10.2.3

himabindu07:
## Details

imageName | oldTag | newTag 
-- | -- | -- 
ap-grafana | 10.0.9 | 10.2.3

## Related Issues

https://github.com/astronomer/issues/issues/5880

## Checklist

[ X ] version.txt was updated


pgvishnuram:
we reverted grafana version due to the permission being revoked on the dashboards - we need validate them before updating to astronomer/astronomer repo

himabindu07:
> we reverted Grafana version due to the permission being revoked on the dashboards - we need to validate them before updating to the astronomer/astronomer repo

If you can provide me with the steps to validate, I can validate

danielhoherd:
@himabindu07 here is the ticket describing the issue. https://github.com/astronomer/issues/issues/5759 the test would be to verify both while logged in and logged out of grafana using admin/admin and see if the same dashboards are available. In grafana 9 some dashboards become unavailable unless you are logged in, and we want to avoid that. we want all dashboards to be available when not logged in.

PR #677, opened 2024-01-16
Update kube-state version 2.10.1

himabindu07:
## Details

imageName | oldTag | newTag 
-- | -- | -- 
ap-kube-state | 2.10.0 | 2.10.1

## Related Issues

https://github.com/astronomer/issues/issues/6080

## Checklist

[x] version.txt was updated


pgvishnuram:
v2.10.0 kube_state_metrics will make changes to labels and annotation after changing this we should verify our metrics in astro-ui 

Label and annotation metrics aren't exposed by default anymore to reduce the memory usage of the default configuration of kube-state-metrics. Before this change, they used to only include the name and namespace of the objects which is not relevant to users not opting in these metrics.

pgvishnuram (approved):

PR #682, opened 2024-01-24
Revert upstream registry repo

pgvishnuram:
## Details

revert upstream registry repo since aws iam role is broken in distribution repo upstream

## Related Issues

https://github.com/astronomer/issues/issues/5939
https://github.com/astronomer/issues/issues/5905

## Checklist

- [x] version.txt was updated


rishkarajgi:
Pre-commit failing

rishkarajgi:
`since aws iam role is broken in distribution repo upstream`

Can you provide more details as to how it is broken? Is this an upstream error or does it require any additional validation on our end?

danielhoherd (commented):
79 is an astonishingly high number of CVEs that we are allowing. Is there no way to build this version of registry but on a newer golang or with newer modules so that we can trim down that list of CVEs?

danielhoherd:
```
Vulnerabilities found for image ap-registry:092ab6cddc8d2906a078a2c101583453a23c5a53: total - 92, critical - 7, high - 52, medium - 27, low - 6
```

I can't imagine that our customers will be OK running this.

pgvishnuram:
currently customers are running old version of registry 3.16 we have not upgraded to 3.18 yet

pgvishnuram:
if we want to upgrade then we might need to add changes releases to aws iam roles - since we have time we can try making it

PR #686, opened 2024-02-05
Add OpenLineage specific statsd mappings

pritt20:
## Details

This change is to add OpenLineage specific statsd mappings to capture OpenLineage metrics.

## Related Issues

https://github.com/astronomer/astro/issues/19117

## Checklist

- [x] version.txt was updated

PR #688, opened 2024-02-08
fix openlineage specific statsd mappings

pritt20:
## Details

This change is in continuation of PR https://github.com/astronomer/ap-vendor/pull/686. After some testing we found out that we would need `airflow` as prefix for openLineage provider statsd mappings as well. Hence, in this change we are fixing these required openLineage statsd mappings by adding `airflow` prefix

## Related Issues

https://github.com/astronomer/astro/issues/19117

## Checklist

- [x] version.txt was updated


danielhoherd (approved):
Seems fine to me, but once again we need somebody from gen2 to sign off. @jpweber @goutamtadi1

ianbuss (approved):

pgvishnuram (approved):

PR #695, opened 2024-03-01
Add Azure Container Registry as an image push target

jpweber:
## Details

This PR will Azure Container Registry (ACR) as a push target as we transition to using that as the primary source for hosting our container images in astro. Not many of these images are used in astro but a few of them are. I decided to just push them all in case software wants to transition someday, but also to not add messy logic for selecting specific repos to push. 

ACR credentials have been added to circle ci under the azurecr.io name.

## Related Issues

None

## Checklist

- [ ] version.txt was updated - NA


danielhoherd:
LGTM

danielhoherd (approved):

PR #727, opened 2024-03-25
Fixing part of the azure validate tags job

jpweber:
## Details

Vishnu reported that this CI job was failing to push images to ACR. I looked in to it and spotted something that was different for the ACR version of the job versus quay and docker. This PR contains what I think is the fix. 

Changed value of image name in validate tags job, within the ACR section, to match the parameter the other registries use. 

## Related Issues

No related issue. Bug reported to me directly via slack.


danielhoherd (approved):
Good catch! Looks good. We need to regenerate the config from the j2 file though either by using `pre-commit run --all-files` or by running `bin/generate_circleci_config.py` and committing the changes.

PR #731, opened 2024-03-26
update uid for init image to run natively on openshift

pgvishnuram:
## Details

update uid for init image to run natively on openshift

## Related Issues

Yet to create

## Checklist

- [x] version.txt was updated


rob-1126 (commented):
Yay!

PR #756, opened 2024-04-08
update dogstatsd to v7.52.1 which has no cves

AmFlint:
## Details

Update Dogstatsd image from v7 tag to more specific v7.52.1 tag which I just scanned with trivy and seems to resolve CVEs we found in the past in production.

## Checklist

- [x] version.txt was updated


danielhoherd (changes requested):
There is a root user issue being reported by Twistlock. We need to somehow override that, or make this not use root user. We will be unable to release the built image if we do not address that twistlock complaint.

pgvishnuram:
if @AmFlint  is okay with that - they can add in twistcliignore and release them

danielhoherd:
@AmFlint if you're OK with ignoring the CVEs then we are good to merge this.

PR #767, opened 2024-04-29
Export more Airflow metrics and add labels

astro-walker:
## Details

Add more Airflow metrics to the `statsd-exporter` mappings to more closely match the [`dogstatsd` configuration](https://github.com/astronomer/ap-vendor/blob/main/dogstatsd/include/dogstatsd.yaml), and label various Airflow metrics with IDs that make the timeseries data meaningful. 

Many Airflow metrics are ~ useless to users without labels to group/filter them on. For example, duration metrics for deployments with > 1 DAG commingle metrics for all DAGs, which makes it impossible to spot regressions in the duration of a single DAG. Similarly, failure counters with no labels prevent users from determining what has failed. This will increase the cardinality of the metrics `statsd-exporter` emits, but to do otherwise would be to put the cart before the horse: some dimensions are needed for the metrics to be useful to users.

There are also some minor whitespace changes to group related metrics into blocks.

## Related Issues
https://github.com/astronomer/astro/issues/20184

## Checklist

- [x] version.txt was updated


statsd-exporter/twistcliignore
```diff
@@ -1,9 +1 @@
-CVE-2023-39318
-CVE-2023-39319
-CVE-2023-39323
-CVE-2023-39326
-CVE-2023-45283
-CVE-2023-45284
-CVE-2023-45285
-CVE-2023-45287
-CVE-2023-48795
+CVE-2023-45288
```
astro-walker, line 1: The scanner reported that all the existing CVEs in the ignore file were resolved and detected this one as a new finding: https://app.circleci.com/pipelines/github/astronomer/ap-vendor/3010/workflows/011ccc93-d2f9-4401-b700-561da8aa0ef6/jobs/50850

We are using the most up to date `statsd-exporter` release, which means that this CVE exists in the upstream image.

statsd-exporter/include/mappings-gen2.yml
```diff
@@ -6,20 +6,27 @@
...
+  - match: "airflow.*_end"
+    name: "airflow_job_end"
+    labels:
+      job_name: "$1"
+  - match: "airflow.*_heartbeat_failure"
+    name: "airflow_job_heartbeat_failure"
     labels:
-      # This is here to make the Astro UI work
-      # because it's expecting an operator label
-      operator: "Value"
```
bote795, line 16: idk if this will break anything, based on the comment above?

astro-walker: It needs a label named `operator`, which still exists and is now populated with the actual operator name.

bote795 (approved):

PR #798, opened 2024-06-24
update alertmanager dockerfile

pgvishnuram:
## Details

Modifies entire alertmanager Dockerfile to support ca certificates chain update based on enable and disable flag 

## Related Issues

https://github.com/astronomer/issues/issues/6334

## Checklist

- [ ] version.txt was updated


alertmanager/Dockerfile
```diff
@@ -14,10 +14,33 @@
 # limitations under the License.
 
 # https://hub.docker.com/r/prom/alertmanager/tags?page=1&ordering=last_updated
-FROM prom/alertmanager:v0.27.0
-LABEL maintainer="Astronomer <humans@astronomer.io>"
+FROM prom/alertmanager:v0.27.0 AS UPSTREAM
 
+FROM quay.io/astronomer/ap-base:3.18.6-2
```
danielhoherd, line 19: @pgvishnuram Why did we change this to a multi-stage build?

pgvishnuram: im unable to find a way to add update-ca-certificates in busybox base image

danielhoherd: That makes sense. This looks like the Dockerfile: https://github.com/prometheus/alertmanager/blob/main/Dockerfile I think using a two-stage build is fine in this case.

alertmanager/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/sh
+
+set -e
+
+# In production, we mount the registry TLS certificate
+# into /usr/local/share/ca-certificates. This completes
+# the certificate installation so we can trust the registry.
+if [ "$UPDATE_CA_CERTS" = "true" ]; then
+echo "Running update-ca-certificates"
+update-ca-certificates
```
danielhoherd, line 10:
```suggestion
  echo "Running update-ca-certificates"
  update-ca-certificates
```

alertmanager/Dockerfile
```diff
@@ -14,10 +14,33 @@
...
+COPY --from=UPSTREAM /bin/alertmanager /usr/local/bin
+COPY --from=UPSTREAM /etc/alertmanager/alertmanager.yml   /etc/alertmanager/alertmanager.yml
+COPY --chown=65534:65534 entrypoint.sh /entrypoint.sh
+
+# Adopt openshift permissions
+RUN chown -R 65534:65534 /etc/ssl/ \
+    && chown -R 65534:0 /etc/ssl/certs \
+    && chmod -R ug+rwx /etc/ssl/certs
+
+
```
danielhoherd, line 38:
```suggestion
# Copied from https://github.com/prometheus/alertmanager/blob/main/Dockerfile
```

danielhoherd (approved):
Seems fine to me. Having tests would be great, but I don't think that should block merging this.

PR #816, opened 2024-07-22
Add state label to Airflow task instance finish metric

astro-walker:
## Details

Retain the `state` label for `airflow.ti.finish` metrics. Without this label, we cannot determine the final state that a task instance ends up in i.e. success or failure.

## Related Issues

https://github.com/astronomer/astro/issues/22692

## Checklist

- [x] version.txt was updated


danielhoherd (approved):
Seems fine to me. This is gen-2 only.

PR #839, opened 2024-08-13
6558 play with modernizing fluentd

danielhoherd:
## Details

This branch unpins all the modules we have in fluentd to see what would happen if we just accepted the latest versions of all modules. Some of these modules are pinned to really old versions, and the reasons for those pins were not documented. In the interest of in breaking the cargo cult, and in order to have something more modern to help diagnose the failure in https://github.com/astronomer/issues/issues/6558, I think we should merge this and see how it goes.

Also notable, the upstream repo is gone so I deleted the comment about where it was from.

## Related Issues

- Tracking issue: https://github.com/astronomer/issues/issues/6558
- PR where the problem was discovered: https://github.com/astronomer/astronomer/pull/2282

## Checklist

- [x] version.txt was updated


danielhoherd:
I went ahead and built this and pushed it to quay.io/danielhoherd/scratch:fluentd-1-17-1-a and then patched it into CI and it's working.

fluentd/include/Gemfile
```diff
@@ -1,28 +1,27 @@
-# The original upstream Gemfile is: https://raw.githubusercontent.com/kubernetes/kubernetes/master/cluster/addons/fluentd-elasticsearch/fluentd-es-image/Gemfile
 # To update, loosely version these packages, then run 'make update-fluentd-gemfile.lock'
 
 source 'https://rubygems.org'
 
 ### Upstream packages
-gem 'activesupport', '~>6.1'
-gem 'elasticsearch', '<7.14'
```
pgvishnuram, line 8: @danielhoherd we pinnned elasticsearch to older version because of the issue with elasticsearch policy previously thats why its pinned to old version

fluentd/include/Gemfile
```diff
@@ -1,28 +1,27 @@
...
-gem 'elasticsearch-xpack', '<7.14'
-gem 'fluent-plugin-concat', '~>2.5'
-gem 'fluent-plugin-detect-exceptions', '~>0.0'
-gem 'fluent-plugin-elasticsearch', '~>4.3'
```
pgvishnuram, line 12: same here

fluentd/include/Gemfile
danielhoherd: The PR for pinning it says it was pinned because of a license concern. https://github.com/astronomer/ap-vendor/pull/184

danielhoherd: @pgvishnuram do we know exactly what that issue is? did we make an ticket describing the problem? do we have a link to a slack thread where we discussed it?

Simpcyclassy: Has this been confirmed?

danielhoherd: No. We have decided to build this version and test it out anyway. We want to see if this giant leap of versions breaks, and if we can find ways to fix anything that breaks.

pgvishnuram (approved):

Simpcyclassy (approved):
OK. LGTM then.

PR #878, opened 2024-10-15
Change Cache time to a week

Shubham-astro:
## Details
This PR intends to increase the TTL for cache that is used by trivy-scan so that we do not get rate-limit exceeded error. 

## Related Issues

Related https://github.com/astronomer/issues/issues/6689


danielhoherd (approved):
CI didn't pass because there's no build, but that's fine.

This uses the iso8601 week format for the cache key.

```
$ date +%Y-W%W
2024-W42
```

PR #886, opened 2024-11-07
fluentd 1.17.1, fix webrick CVE

danielhoherd:
## Details

Rebuild fluentd requirements, which solves webrick CVE and supersedes https://github.com/astronomer/ap-vendor/pull/885

## Related Issues

NA, KTLO work

## Checklist

- [x] version.txt was updated


Shubham-astro (approved):
Looks good, simple version upgrade

Shubham-astro:
@danielhoherd I see scan-twistlock  has failed due to:
```
(CIS_Docker_v1.5.0 - 4.1) Image should be created with a non-root user 
```

I think this may need a separate ticket to resolve.?

danielhoherd:
@Shubham-astro fluentd needs root permissions. It is one of the only components in our stack that needs it. Fluentd must mount bare-metal log directories that are only accessible to the bare-metal OSes root user.

pgvishnuram:
Fluentd requires root privileges to read node logs we cannot remove it

danielhoherd:
The problem version of ruby is installed as part of the core ubuntu 24.04 ruby libraries:

```
# dpkg -S /usr/lib/ruby/gems/3.2.0/gems/rexml-3.2.5
libruby3.2:arm64: /usr/lib/ruby/gems/3.2.0/gems/rexml-3.2.5
```

This package is currently `3.2.3-1ubuntu0.24.04.3`, so we'll have to wait for a new version of that.

danielhoherd:
11 days later ubuntu still has not patched the libruby CVEs here.

PR #941, opened 2025-02-06
6786 Move git-sync-relay build to github.com/astronomer/git-sync-relay (private repo)

danielhoherd:
## Details

I am moving the git-sync-relay build to a private repo. This is for a few reasons:
- The rewrite of the component makes this something that is not FLOSS code anymore.
- The build requires accessing a private git repo, which ap-vendor is not set up to do.
- This keeps all the build components in a single repo, not spread between the python code repo and the docker image build repo.
- This isn't vendored software, it's our software, so this really isn't the right repo for it.

## Related Issues

<https://github.com/astronomer/issues/issues/6786>

## Checklist

- [X] version.txt was deleted 😆

Do not merge this until we have a proper build running from https://github.com/astronomer/git-sync-relay


pgvishnuram:
if we remove this for the older versions of platform how we be able to maintain patch fixes

danielhoherd:
> if we remove this for the older versions of platform how we be able to maintain patch fixes

@pgvishnuram we will patch them in the new repo. The current version is what we want all future git-sync deploys to use, because it adds the features requested in https://github.com/astronomer/issues/issues/6797

PR #951, opened 2025-03-04
add signal handlers to sidecar_server.py script

Shubham-astro:
## Details

This PR adds proper signal handling to the `sidecar_server.py` script to ensure container graceful termination. Currently, when a kill signal (like SIGTERM) is sent to the Python process , it doesn't properly propagate to the vector process, causing containers to hang during shutdown and requiring manual intervention.

### Problem
- The vector process doesn't receive termination signals when they are sent to the parent Python process 
- This causes containers to get stuck during shutdown, failing to terminate cleanly
- The issue is caused by using `shell=True` in subprocess creation, which creates an intermediate shell process

### Solution
1. Removed `shell=True` from the subprocess.Popen call to eliminate the intermediate shell
2. Added explicit signal handlers for SIGTERM and SIGINT signals
3. Ensured proper signal propagation to the vector process
4. Improved the quit_proc method to ensure all child processes are terminated

### Impact
- Containers will now terminate gracefully when receiving SIGTERM signals
- Reduces hanging containers in Kubernetes environments
- Improves overall system reliability and resource management

## Related Issues

https://github.com/astronomer/issues/issues/6714

## Checklist

- [ ] version.txt was updated

## Risk Assessment

**Low Risk**: This change affects container termination but doesn't impact the core functionality of the application. The risk is limited to the shutdown process, which is now more reliable.

## Testing

- Manually verified signal propagation on development cluster

Before the change the kill was not working:
Screenshot 2025-03-04 at 1 17 24 PM

After the change the kill is working as expected when the sidecar script receives the signal:
Screenshot 2025-03-04 at 1 17 58 PM


vector/sidecar_server.py
```diff
@@ -15,21 +15,28 @@
...
 from pathlib import Path
 
-ppid = os.getppid()
-print(f"{ppid=}", flush=True)
+
+def signal_handler(sig):
+    print(f"Received signal {sig}", flush=True)
+    if handler:
+        handler.quit_proc()
+    sys.exit(0)
```
danielhoherd, line 29: when testing this, after running `kill 1` i saw this in the vector logs:

```
Traceback (most recent call last):
  File "/usr/local/bin/sidecar_server.py", line 86, in <module>
    handler.run()
  File "/usr/local/bin/sidecar_server.py", line 76, in run
    time.sleep(5)
TypeError: signal_handler() takes 1 positional argument but 2 were given
```

vector/sidecar_server.py
```diff
@@ -15,21 +15,28 @@
...
+import signal
 import subprocess
+import sys
 import time
 from pathlib import Path
 
-ppid = os.getppid()
-print(f"{ppid=}", flush=True)
+
+def signal_handler(sig):
```
danielhoherd, line 25: This fixes the `signal_handler() takes 1 positional argument but 2 were given` error:
```suggestion
def signal_handler(sig, _):
```

danielhoherd (approved):
Once my suggestion is merged this is good to go.

PR #965, opened 2025-03-10
Upgrade to ap-base:3.21.3-1

danielhoherd:
## Details

- Upgrade to ap-base:3.21.3-1, which is a jump from alpine 3.20 to 3.21
- Upgrade some build tool versions (orbs, ci runners, pre-commit hooks)
- Update some component versions to solve CVEs

## Related Issues

- <https://github.com/astronomer/issues/issues/6830>

## Checklist

- [x] version.txt was updated


nats-exporter/Dockerfile
```diff
@@ -14,23 +14,23 @@
 # limitations under the License.
 
 # Golang binary building stage
-ARG GO_VERSION=1.21
+ARG GO_VERSION=1.23
```
pgvishnuram, line 17: the underlying golang version for nats exporter for 1.15 is 1.21 https://github.com/nats-io/prometheus-nats-exporter/blob/v0.15.0/go.mod#L3

danielhoherd: It builds fine on 1.23 though. Is 1.23 going to cause problems?

danielhoherd: nats-exporter 0.16.0 uses golang 1.23, so i'm going to bump it to that. https://github.com/nats-io/prometheus-nats-exporter/releases/tag/v0.16.0

PR #966, opened 2025-03-12
Parse bundle backend metrics in statsd-exporter

jeremybeard:
## Details

This changes the statsd-exporter metrics mappings to include those emitted by the new Astro bundle backend.

## Related Issues

- https://github.com/astronomer/astro-bundle-backend/issues/13
- https://github.com/astronomer/astro-bundle-backend/pull/16

## Checklist

- [ ] version.txt was updated


jeremybeard:
@pgvishnuram Do I need to bump the patch version in `version.txt` even though I haven't bumped the underlying `statsd-exporter` version? Also I think your CVE push has linting errors.

danielhoherd:
@jeremybeard yes. version.txt controls which version is used for the docker tag. We never overwrite existing tags, so we always need to bump version.txt.

jeremybeard:
@danielhoherd Thanks, I just bumped it (down) to 0.28.0-2 because #952 is actually using 0.28.0-1 (the base image is 0.28.0)

PR #991, opened 2025-04-05
add trivy validator for ap-vendor

pgvishnuram:
## Details

Do not merge this PR until this description is replaced with relevant info about what images are being updated and what changes are being made.

## Related Issues

## Checklist

- [ ] version.txt was updated


bin/trivy_validator.py
```diff
@@ -0,0 +1,40 @@
...
+    if sorted_items:
+        return "\n\n" + "\n".join([f"{item}" for item in sorted_items]) + "\n"
+    return "none\n"
+
+
+def get_ignored_cves() -> list:
+    ignore_file = Path(f"{scan_results_file_path}/trivyignore")
+    if ignore_file.is_file():
+        return sorted({line.strip() for line in ignore_file.read_text().splitlines()})
+    return []
```
pgvishnuram, line 25: i can import this same function from twist_cli validator python script since both share the same code

pgvishnuram:
@danielhoherd

bin/trivy-scan.sh
```diff
@@ -38,4 +38,6 @@ if grep -q -i 'OS is not detected' trivy-output.txt ; then
   exit 0
 fi
 
-exit "${exit_code}"
+#exit "${exit_code}"
+export PROJECT_DIRECTORY="${scan_target}"
+python bin/trivy_validator.py
```
danielhoherd, line 43: I see that trivy_validator.py exits non-zero if there are CVEs, but I think we should keep this logic in as a fallback. It doesn't hurt.

```suggestion
export PROJECT_DIRECTORY="${scan_target}"
python bin/trivy_validator.py
exit "${exit_code}"
```

bin/trivy_validator.py
danielhoherd: I don't think doing the code changes necessary to make that work are important. It's more important that this script be self contained.

bin/trivy_validator.py
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env python3
+import json
+import os
+from pathlib import Path
+
+git_root = [x for x in Path(__file__).resolve().parents if (x / ".git").is_dir()][-1]
```
danielhoherd, line 6: The best version I can find of this assignment is:

```suggestion
git_root = next(iter([x for x in Path(__file__).resolve().parents if (x / ".git").is_dir()]), None)
```

PR #1050, opened 2025-08-05
remove twistcli config

pgvishnuram:
## Details

remove twistcli ignore from ap-vendor since we longer hold the subscription 

## Related Issues

## Checklist

- [ ] version.txt was updated


danielhoherd (approved):
LGTM.

PR #1057, opened 2025-08-21
Add tini to the alpine image

nickstenning:
## Details

Adds `tini` to the alpine image.

Having access to tini in a base image like this makes it convenient to use the base image for utility containers that don't justify an image of their own.

Specifically, a container that just runs a shell script with `/bin/sh -c ...` will usually have the problem that it doesn't forward SIGTERM (from Kubernetes) to child processes.

With tini we can make such containers run `/sbin/tini -- /bin/sh -c ...` and thus behave correctly in Kubernetes.

## Related Issues

This will be useful for https://github.com/astronomer/astro/issues/33755

## Checklist

- [x] version.txt was updated


nickstenning:
I will totally understand if on principle we don't want to add any dependencies to this image. I just thought I'd open this to see what folks thought, because `tini` is such a special case and it seems like it would be very handy to have this base image with `tini` in it for running simple containers that just run a shell script.

pgvishnuram (approved):

nickstenning:
Thanks @danielhoherd!

PR #1223, opened 2026-05-07
Add statsd mapping for astro_event_scheduler metrics

dstandish:
## Summary
- Add a regex mapping in `statsd-exporter/include/mappings-gen2.yml` that matches `airflow.astro_event_scheduler.*` and exports them as `astro_event_scheduler_*`, stripping the `airflow` prefix.

## Test plan
- [ ] Verify metrics emitted under `airflow.astro_event_scheduler.<name>` appear as `astro_event_scheduler_<name>` after passing through statsd-exporter.
- [ ] Confirm the catch-all drop rule below still drops unrelated metrics.

Contributes to https://linear.app/astronomer/issue/ACD-102/capture-event-scheduler-component-metrics-and-ship-to-astro

🤖 Generated with [Claude Code](https://claude.com/claude-code)


statsd-exporter/include/mappings-gen2.yml
```diff
@@ -291,6 +291,11 @@ mappings:
       instance: "$1"
       mount_path: "$2"
 
+  # Astro Event Scheduler metrics — accept all and strip the airflow prefix
+  - match: 'airflow\.astro_event_scheduler\.(.+)'
```
neel-astro, line 295: minor nit: if we plan to expose per dag or task metrics, then we would want to skip those to keep the metrics at lower cardanality

neel-astro (approved):
LGTM, left a comment around cardinality

rishkarajgi (approved):
Adds a regex mapping for `airflow.astro_event_scheduler.*` metrics that strips the `airflow.` prefix before the new rule hits the catch-all drop below it. The regex and template syntax look correct for the statsd-exporter config format.

**Verdict:** ✅ Looks good — minimal, well-placed change.

statsd-exporter/include/mappings-gen2.yml
dstandish: ok -- for now we don't have high cardinality metrics so i think we're ok to proceed. but i'll add a note in our claude.md about this

neel-astro (approved):

PR #1229, opened 2026-05-12
statsd-exporter: histogram mapping for astro_event_scheduler SLA metrics

dstandish:
Linear: [ACD-103](https://linear.app/astronomer/issue/ACD-103/capture-and-ship-sla-metrics-to-chronosphere)

## Summary

Two commits on this branch:

1. `Bump statsd-exporter version to 0.29.0-3` — picks up upstream changes needed for `match_metric_type: timer` + histogram bucket support.
2. `Add statsd mapping for astro_event_scheduler metrics` — earlier catch-all rule for `airflow.astro_event_scheduler.*` counters and gauges.
3. (this PR's main change) — adds specific timer→histogram rules with sub-second-focused buckets for the two SLA metrics:

   - `astro_event_scheduler_sla_time_to_first_task`
   - `astro_event_scheduler_sla_task_lag`

   Buckets: `[0.1, 0.25, 0.5, 0.75, 1, 2, 5, 10, 30, 60]`. The boundaries cluster resolution around the sub-second SLO line and extend out to 60s to capture stuck-task outliers.

The new rules must precede the existing catch-all `airflow\.astro_event_scheduler\.(.+)` match — that catch-all would otherwise turn the dotted suffix into `astro_event_scheduler_sla.time_to_first_task` (literal dot, invalid in Prometheus) and wouldn't apply histogram bucketing.

## Companion PRs

- [astronomer/astro#38556](https://github.com/astronomer/astro/pull/38556) — OTel relay rename, allowlist, and label drops.
- [astronomer/astro-event-scheduler#145](https://github.com/astronomer/astro-event-scheduler/pull/145) — the in-app emission.

## Test plan

- [ ] In a staging environment, trigger a DAG that completes, port-forward the airflow statsd-exporter, confirm `astro_event_scheduler_sla_time_to_first_task_bucket{le="..."}` series appear with the new boundaries.
- [ ] Confirm `_sum` and `_count` series are present (histogram, not summary).
- [ ] Confirm existing `astro_event_scheduler_events_received` / `loop_blocked` / `active_schedulers` still flow (the catch-all rule below should keep handling them).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


statsd-exporter/include/mappings-gen2.yml
```diff
@@ -291,6 +291,29 @@ mappings:
...
+    name: "astro_event_scheduler_sla_time_to_first_task"
+    timer_type: histogram
+    buckets: [0.1, 0.25, 0.5, 0.75, 1, 2, 5, 10, 30, 60]
+  - match: 'airflow\.astro_event_scheduler\.sla\.task_lag'
+    match_type: regex
+    match_metric_type: timer
+    name: "astro_event_scheduler_sla_task_lag"
+    timer_type: histogram
+    buckets: [0.1, 0.25, 0.5, 0.75, 1, 2, 5, 10, 30, 60]
+
```
rishkarajgi, line 316: If `task_lag` can grow beyond a minute in practice (e.g. for deployments where the scheduler is backpressured or the worker queue drains slowly), anything above 60s lands in `+Inf` and you lose all resolution in that tail. Worth checking the p99 against prod data before this ships — might want to extend to 120 or 300.

statsd-exporter/include/mappings-gen2.yml
```diff
@@ -291,6 +291,29 @@ mappings:
...
+  # Astro Event Scheduler metrics — accept all and strip the airflow prefix
+  - match: 'airflow\.astro_event_scheduler\.(.+)'
+    match_type: regex
```
rishkarajgi, line 331: Nit: if the `astro_event_scheduler.sla.*` namespace ever grows a new metric that isn't listed above, it'll fall through to this catch-all and produce a Prometheus name with a literal dot in it (e.g. `astro_event_scheduler_sla.new_metric`). Could add a third rule here for `airflow\.astro_event_scheduler\.sla\.(.+)` → `astro_event_scheduler_sla_${1}` as a safety net, even if there's nothing concrete planned today.

rishkarajgi (approved):
Bumps statsd-exporter to 0.29.0-3 (needed for `match_metric_type: timer` + histogram support) and adds two explicit histogram mappings for the `astro_event_scheduler` SLA metrics, plus a catch-all to strip the `airflow.` prefix from any remaining scheduler metrics.

**Verdict:** ✅ Looks good — ordering is correct and the comment explaining why specific rules must precede the catch-all is helpful.

statsd-exporter/include/mappings-gen2.yml
```diff
@@ -291,6 +291,29 @@ mappings:
...
+    name: "astro_event_scheduler_${1}"
```
rishkarajgi, line 332: The catch-all `airflow\.astro_event_scheduler\.(.+)` rule doesn't specify `match_metric_type`, so any future SLA timer metric that doesn't have a dedicated rule above will fall through here and be exported as a summary (statsd-exporter's default for timers) rather than a histogram. Might be worth a comment noting this, or considering whether we want a second catch-all scoped to `match_metric_type: timer` with a default histogram config — though that could also make the mapping harder to read. At minimum, a note in the comment block above the catch-all would help future contributors know to add explicit timer rules.

rishkarajgi (commented):
Adds specific timer→histogram rules for `astro_event_scheduler` SLA metrics ahead of the existing catch-all, and bumps statsd-exporter to 0.29.0-3 for `match_metric_type: timer` + histogram bucket support.

**Verdict:** ✅ Looks good — the rule ordering and bucket choices are sensible. One note below about the catch-all's behavior for future timer metrics.

statsd-exporter/include/mappings-gen2.yml
dstandish: Going to leave the buckets as-is for v1 and re-tune from prod data, for a few reasons:

- These buckets are explicitly tuned for sub-second SLO observability. Once `task_lag` exceeds 60s, the deployment is so far outside its SLO that the exact magnitude in the +Inf bucket is less actionable than the SLO breach itself — we'll see it via the +Inf bucket counts, just not the precise distribution.
- More buckets = more series per histogram per deployment. Going to 120 + 300 adds two more `le` values, ~20% bump in histogram series. Not huge, but not free given we're shipping this to every LL-enabled deployment.
- The README ([packages/astro-event-scheduler/README.md](https://github.com/astronomer/astro-event-scheduler/blob/main/packages/astro-event-scheduler/README.md)) explicitly calls out re-tuning from the production histogram shape before changing the boundaries — once we have a week or two of real data we'll know whether the +Inf bucket is actually getting traffic.

If +Inf turns out to be hot in prod, easy follow-up PR to extend.

statsd-exporter/include/mappings-gen2.yml
dstandish: Good catch — applied in ba65b0ee. Added an `airflow\.astro_event_scheduler\.sla\.(.+)` rule between the explicit SLA rules and the broader catch-all so any future sla.* metric without a dedicated mapping still gets a clean Prometheus name.

statsd-exporter/include/mappings-gen2.yml
dstandish: Added a comment to the catch-all in ba65b0ee noting that future timer metrics falling through will be exported as summaries and need an explicit histogram rule above. Going with the comment-only approach rather than a timer-scoped catch-all — a default-bucketed histogram for unknown timers feels worse than a clear "add a rule" instruction, since bucket choice depends heavily on the metric's expected range.

pgvishnuram (approved):

pgvishnuram:
this wont affect APC since this is astro specific and this rules are used only by astro team

dstandish:
@pgvishnuram I had to rebase and update the tag -- you want to double check? CI also needs an approval

pgvishnuram (approved):

dstandish:
@pgvishnuram @rishkarajgi can you guys merge this? i don't have the power

PR #1231, opened 2026-05-13
APC-1298: Package updates for CVE remediation

Shubham-astro:
## Details

This PR remediates by rebuilding chainguard based images. 
CVE likely addressed across these images:
- CVE-2026-33186
- CVE-2026-33729
- CVE-2026-27143

## Related Issues

https://linear.app/astronomer/issue/APC-1298/fix-112-cves

## Checklist

- [x] version.txt was updated


rishkarajgi (approved):

registry/Dockerfile
```diff
@@ -15,7 +15,7 @@
 
 # https://console.chainguard.dev/org/astronomer.io/images/organization/image/ap-registry/versions
 # Registry image uses sha256 digest to ensure we are using the exact same image as the upstream, which is important for security and stability. The version of the registry image can be found in version.txt, but the sha256 digest must be looked up manually on chainguard.dev.
-FROM cgr.dev/astronomer.io/ap-registry@sha256:18f1cb5c3fa1b36861c67f668a50ec6c748d02e8006374716d9a43a0ba8d58fe AS upstream
+FROM cgr.dev/astronomer.io/ap-registry:3.1.1 AS upstream
```
rishkarajgi, line 19: The comment just above says the sha256 digest is intentionally used here for security and stability, and that the sha256 needs to be looked up manually on chainguard.dev. This change switches to a tag (`3.1.1`) instead — tags are mutable, so a rebuild could theoretically pull a different image if the tag is ever re-pointed. Should either update the comment to reflect this is now tag-based, or look up and pin the sha256 for `3.1.1`.

rishkarajgi (commented):
Bumps build versions for grafana, kube-state, prometheus, and statsd-exporter (patch build increments for CVE remediation), and upgrades the registry image from `3.0.0-10` to `3.1.1`.

**Verdict:** 💬 A few questions — the Dockerfile change swaps the pinned sha256 digest for a tag, which is worth addressing given the existing comment about why sha256 is used.

PR #1237, opened 2026-05-21
Update Vector to 0.55.0

Shubham-astro:
## Details

Update vector to use 0.55

## Checklist

- [x] version.txt was updated


vector/version.txt
```diff
@@ -1 +1 @@
-0.54.0-1
+0.55.0-1
```
sudarshanrampuria, line 1:
```suggestion
0.55.0
```

sudarshanrampuria (approved):

vector/version.txt
Shubham-astro: This won't work as we already have 0.55.0, we are rebuilding 0.55.0 to get latest sha and CVE fixes, therefore 0.55.0-1

PR #1248, opened 2026-05-22
skip-cve-for-which-fix-version-is-not-present

sudarshanrampuria:
## Summary

CI now only fails on critical/high CVEs that have a fix version available.
CVEs without a fix version are still printed in a separate informational table so they remain visible.
Fixed a finadings typo that broke the script.

## Testing
image


rishkarajgi (approved):

rishkarajgi (changes requested):
Holding until we verify with the security team if Endor scans upstream Chainguard for fix available

rishkarajgi (approved):

PR #1296, opened 2026-06-05
Upgrade Datadogstatsd + non-root user

AmFlint:
## Details

In this Pull Request, I have updated the datadog-statsd base image to use the latest available version, and I have also made some changes in the Dockerfile to make it possible to run this container as non-root user, in more security-sensitive environments, with a specific securityContext that would enforce running as non-root.

Related to https://linear.app/astronomer/issue/AINF-1993/update-ap-dogstatsd-image-to-run-as-non-root

## Related Issues

## Checklist

- [x] version.txt was updated


jpweber (approved):

ianbuss (approved):
Assuming the CVEs are not genuine things we need to fix.

dogstatsd/test.yaml
```diff
@@ -1,5 +1,11 @@
...
-  root_user_test: False
+  root_user_test: True
+  default_user: datadog
+  users_config:
+    - name: datadog
+      group: datadog
+      gid: 1000
+      uid: 1000
 docker:
   entrypoint: "/bin/sleep 1m"
```
amccabe-astronomer, line 5: a worry is dogstatsd/Dockerfile ln 34-38 might break in a way the entrypoint here wouldn't catch, it

amccabe-astronomer: I can pass it, but mentioning it in case you have ideas/otherwise have checked it.

dogstatsd/Dockerfile
```diff
@@ -14,11 +14,29 @@
...
 LABEL maintainer="Astronomer <humans@astronomer.io>"
 
 ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 
-COPY --chown=root:root include/dogstatsd.yaml /etc/datadog-agent/dogstatsd.yaml
+USER root
+
+# The upstream image runs as root. Create an unprivileged "datadog" user and
```
amccabe-astronomer: I usually review then run a local review in claude with something like "please do antagnostic code review on <pr> for security and correctness" to check my work.

I am not sure about the details on this, but passing it along

```
  2. The Dockerfile comment is likely wrong (correctness/maintainability)

  The comment justifies chmod +x /dogstatsd and the chown of /dogstatsd with:

  ▎ /dogstatsd : the entrypoint runs chmod +x on it, which only succeeds if the running user owns the file
  
  I pulled the upstream alpine entrypoint.sh at both main and tag 7.79.2. It does none of that — it only validates DD_API_KEY, conditionally sets DD_DOGSTATSD_NON_LOCAL_TRAFFIC, and exec "$@". There is no runtime chmod. So the stated rationale
  is unsupported. The chmod +x in the build layer is harmless, but a wrong "why" comment is worse than none — the next person will trust it. Either point to the real source of the chmod behavior (a non-alpine variant? an init wrapper?) or
  delete the claim. Same scrutiny applies to the "auth_token + ipc_cert.pem are written here on startup" line — plausible for 7.79's IPC auth, but it's asserted, not demonstrated.
```

dogstatsd/Dockerfile
```diff
@@ -13,12 +13,12 @@
...
+# https://images.chainguard.dev/directory/image/dogstatsd/versions
+FROM cgr.dev/astronomer.op/dogstatsd:7.78.4
 LABEL maintainer="Astronomer <humans@astronomer.io>"
 
 ARG BUILD_NUMBER=-1
 LABEL io.astronomer.docker=true
 LABEL io.astronomer.docker.build.number=$BUILD_NUMBER
 
-COPY --chown=root:root include/dogstatsd.yaml /etc/datadog-agent/dogstatsd.yaml
+COPY include/dogstatsd.yaml /etc/datadog-agent/dogstatsd.yaml
```
AmFlint, line 24: might need to include a chown here, although I don't know what user/group to chown to, so we'd ideally need to build the image and test it

PR #1332, opened 2026-06-15
Fix alertmanager not found error

sudarshanrampuria:
## Details

Cause: [ap-base #121](https://github.com/astronomer/ap-base/pull/121) removed build-base, deleting the `/usr/local/bin` directory. With the dir gone, COPY <binary> /usr/local/bin treated the destination as a filename, so amtool and alertmanager collapsed onto one file → exec: alertmanager: not found (exit 127).

## Related Issues

## Checklist

- [x] version.txt was updated


Shubham-astro (approved):

alertmanager/endorignore
```diff
@@ -1,3 +1,4 @@
+CVE-2026-45447
```
Shubham-astro: is this CVE present on chainguard as well.?

sudarshanrampuria: It shows it was scanned 5 days ago with no vulnerability.
image

sudarshanrampuria: Seems like this was in ap-base image. Bumped it and the CVE is gone. Nice, we actually caught a CVE using endor scans for the first time 👏

PR #1353, opened 2026-06-22
remove astrocr from vendor mirror list

pgvishnuram:
## Details

remove astrocr from vendor mirror list

## Related Issues

NA

## Checklist

- [ ] version.txt was updated


ianbuss:
Agree we should do this as these are by definition meant to be publicly accessible images but let's hold off merging until we ensure that the `astro` monorepo updates its references to `__public-image-repository__` instead of `__system-image-repository__` for these.

danielhoherd (changes requested):
We will also need to remove the `azurecr.io` context

pgvishnuram:
> We will also need to remove the `azurecr.io` context

we still need them azurpubliccr is still used for azure mirror

Shubham-astro (approved):

PR #1373, opened 2026-06-26
PLAT-564 Enable a shared namespace for shared telemetryfrom within airflow

namratachaudhary:
## Details

This PR is an extension of this change in astro : https://github.com/astronomer/astro/pull/39465
 
While statsd cannot emit histograms in any form, we would still like to have counters / gauges etc that can be emitted with appropriate prefix be collected from both pathways. 

## Related Issues

PLAT-564

## Checklist

- [x] version.txt was updated


claassen (approved):

amccabe-astronomer (approved):
Not a codeowner, but approved!

danielhoherd (approved):
CVE exception list is pretty long, but none of them are solvable at this time.

PR #1474, opened 2026-08-04
Bump-ap-base-in-eso-image

sudarshanrampuria:
## Details

* Bump ap-base version
* Change version to -<0,9> format. Because we are following that for all of our images.


danielhoherd (approved):
Looks good to me. The version string change was discussed in slack. Dropping `astro` brings this version string in line with what we're using elsewhere in ap-vendor.

PR #1511, opened 2026-08-18
PINF-347: Regenerate postgresql.conf/pg_hba.conf via initdb on upgrade, instead of failing to start

Shubham-astro:
## Details

Upgrading in-cluster PostgreSQL (0.37.7 → 2.x) hits a crash even after the socket-directory permission bug  is resolved:
```
PostgreSQL Database directory appears to contain a database; Skipping initialization
postgres: could not access the server configuration file "/bitnami/postgresql/data/postgresql.conf": No such file or directory
```
Inspecting the PVC directly (via a kubectl debug --copy-to clone against a live upgrade test) showed a genuinely valid, previously-initialized PostgreSQL 17 data directory — PG_VERSION, base/, global/, pg_wal/, postmaster.opts, pg_ident.conf, postgresql.auto.conf all present and correctly owned — but postgresql.conf and pg_hba.conf were missing entirely, not even as broken symlinks.

Root cause, confirmed by reading cgr.dev/astronomer.io/ap-postgres's own /usr/bin/docker-entrypoint.sh directly (it's the standard docker-library/postgres entrypoint, rebuilt on Chainguard's base — readable, not opaque as originally assumed): its docker_init_database_dir()/pg_setup_hba_conf() functions only write postgresql.conf/pg_hba.conf the first time they initialize a data directory (gated on PG_VERSION not existing yet), and silently skip that step on every upgrade, where PG_VERSION is already present. Fresh installs work today because they always take the first-init path; upgrades never do.

Fix: docker-entrypoint-wrapper.sh runs before the real entrypoint and repairs only this exact condition (PG_VERSION present, and postgresql.conf/pg_hba.conf missing). Rather than shipping static copies of these files (an earlier version of this fix pulled them once from a live pod — see git history), it now runs this same image's own initdb against a throwaway scratch directory and lifts the two generated files out of it. That means the repair always reflects whatever the currently running image's initdb actually produces, instead of a snapshot that would silently drift as the base image changes. Two things initdb's own output doesn't cover, both handled explicitly:

- initdb looks up its own effective UID via getpwuid() regardless of --username, and Kubernetes' runAsUser doesn't add a matching /etc/passwd entry for it. Worked around with the identical nss_wrapper technique the real entrypoint uses for its own initdb call (docker_init_database_dir()) — LD_PRELOADing libnss_wrapper.so with a synthetic passwd/group entry for the running UID.
- initdb never writes a remote (non-loopback) pg_hba.conf rule on its own — the real entrypoint's pg_setup_hba_conf() appends that separately, based on the server's current password_encryption default. Since this is a repair of a pre-existing install rather than a fresh one, today's default doesn't tell us what an existing role's stored password was actually hashed with — so this appends host all all all md5 explicitly instead: md5 authenticates SCRAM-hashed passwords exactly the same way, but also covers any role whose password predates Postgres's SCRAM default and was never rotated, which scram-sha-256 would otherwise reject outright even though Postgres itself now starts.

## Related Issues

https://linear.app/astronomer/issue/PINF-347/add-readonlyrootfilesystem-volume-mounts-for-internal-postgresql

## Checklist

- [x] version.txt was updated


postgresql/pg_hba.conf
```diff
@@ -0,0 +1,128 @@
...
+# CAUTION: Configuring the system for local "trust" authentication
+# allows any local user to connect as any PostgreSQL user, including
+# the database superuser.  If you do not trust all your local users,
+# use another authentication method.
+
+
+# TYPE  DATABASE        USER            ADDRESS                 METHOD
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     trust
```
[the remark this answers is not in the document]
Shubham-astro: Accurate, but pre-existing — this is the literal config the chainguard base image's own entrypoint already writes for every fresh install; I copied it from there rather than guess at a hardened version, specifically to avoid config drift. Checked the statefulset: only the postgres container mounts the postgresql-run socket, so this doesn't extend trust to the metrics sidecar or anything else in the pod.

postgresql/pg_hba.conf
```diff
@@ -0,0 +1,128 @@
...
+host    all             all             127.0.0.1/32            trust
+# IPv6 local connections:
+host    all             all             ::1/128                 trust
+# Allow replication connections from localhost, by a user with the
+# replication privilege.
+local   replication     all                                     trust
+host    replication     all             127.0.0.1/32            trust
+host    replication     all             ::1/128                 trust
+
+host all all all scram-sha-256
```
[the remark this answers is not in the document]
Shubham-astro: Also pre-existing,  chart doesn't configure TLS anywhere today so adding hostssl here would break existing connections generally, not just the repaired-upgrade case.

postgresql/pg_hba.conf
```diff
@@ -0,0 +1,128 @@
...
+
+# "local" is for Unix domain socket connections only
+local   all             all                                     trust
+# IPv4 local connections:
+host    all             all             127.0.0.1/32            trust
+# IPv6 local connections:
+host    all             all             ::1/128                 trust
+# Allow replication connections from localhost, by a user with the
+# replication privilege.
+local   replication     all                                     trust
```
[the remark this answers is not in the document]
Shubham-astro: But it's pre-existing, not something this PR adds: it's part of the same byte-for-byte config pulled from a live healthy pod running this image, so it's the exact rule every fresh install of this chart already ships with today..

postgresql/pg_hba.conf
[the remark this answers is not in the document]
Shubham-astro: This isn't valid — checked against the PostgreSQL docs directly (auth-pg-hba-conf.html): all is an explicitly documented keyword for the address field, not just database/user ("You can also write all to match any IP address..."). It's equivalent to your suggested 0.0.0.0/0 + ::0/0 split, just shorthand. This exact line also came verbatim from a live pod running this same PostgreSQL build and image tag, which is parsing and serving connections on it right now — that's direct empirical proof it's not a parse failure.

postgresql/pg_hba.conf
```diff
@@ -0,0 +1,128 @@
...
+host    replication     all             127.0.0.1/32            trust
```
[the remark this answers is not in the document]
Shubham-astro: Accurate, and worth noting this one is scoped a bit differently than the local rules discussed elsewhere in this file: 127.0.0.1 loopback is shared across every container in the pod's network namespace, not just whichever container mounts the postgresql-run socket — so this is reachable by any container in the pod, not only the postgres one. Still pre-existing, not introduced by this PR — it's part of the same byte-for-byte config pulled from a live healthy pod.

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,36 @@
...
+
+if [ -f "$data_dir/PG_VERSION" ] && [ ! -f "$data_dir/postgresql.conf" ] && [ -f "$staging_dir/postgresql.conf" ]; then
+  cp "$staging_dir/postgresql.conf" "$data_dir/postgresql.conf"
+fi
+
+if [ -f "$data_dir/PG_VERSION" ] && [ ! -f "$data_dir/pg_hba.conf" ] && [ -f "$staging_dir/pg_hba.conf" ]; then
+  cp "$staging_dir/pg_hba.conf" "$data_dir/pg_hba.conf"
+fi
+
+exec /usr/bin/docker-entrypoint.sh "$@"
```
rishkarajgi, line 68: This is the main thing I'd want confirmed before merge: the whole fix hinges on `/usr/bin/docker-entrypoint.sh` actually being the entrypoint of `cgr.dev/astronomer.io/ap-postgres:17.10`. The Dockerfile comment asserts it, but if the path is wrong every Postgres pod fails to start — a worse outage than the missing-conf bug. Can we confirm it against the real image, and ideally add a boot smoke test so a future base-image change can't silently break the exec?

postgresql/pg_hba.conf
rishkarajgi: Auth posture here is correct — remote is password-based scram-sha-256 (not trust), and the broad address is the normal in-cluster posture scoped by the NetworkPolicy. One thing to confirm for the upgrade path this repairs: were the existing installs' passwords stored as scram hashes? If a pre-Chainguard upgrade's credentials are md5-hashed, a scram-sha-256 rule will reject them and clients won't authenticate even though Postgres now starts.

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,36 @@
...
+#     already exist, nothing is copied, even if their content has since diverged
+#     from this default (e.g. via ALTER SYSTEM, which writes to
+#     postgresql.auto.conf and is unaffected either way).
+#   - It only ever fires once per data directory: once the file exists, every
+#     later start sees the condition as false.
+
+data_dir="${PGDATA:-/bitnami/postgresql/data}"
+staging_dir="/opt/astronomer/conf"
+
+if [ -f "$data_dir/PG_VERSION" ] && [ ! -f "$data_dir/postgresql.conf" ] && [ -f "$staging_dir/postgresql.conf" ]; then
```
rishkarajgi: Nicely guarded — writes only when the data dir is initialized (`PG_VERSION` present) and the conf is genuinely missing, so a healthy install's tuned config is never clobbered and it's a no-op after the first repair. Worth a test that exercises both this repair branch and the healthy-install no-op.

rishkarajgi (changes requested):
Requesting changes — to be clear up front, this isn't because the code looks wrong. The security-critical parts are actually correct (see below). It's that two load-bearing premises can't be verified from this repo, and if either is wrong the failure mode is a fleet-wide Postgres outage or a post-upgrade auth failure — so for the image every internal Postgres runs on, I'd want them confirmed (ideally with a boot smoke test) before merge rather than after.

The two premises to confirm before merge:
1. **The hardcoded `exec /usr/bin/docker-entrypoint.sh "$@"` must actually be the base image's entrypoint.** The Dockerfile and wrapper both assert this about `cgr.dev/astronomer.io/ap-postgres:17.10`, but it isn't verifiable from this repo — and if that path is wrong, every Postgres pod fails to start, which is worse than the bug being fixed. A boot smoke test (container starts, Postgres comes up, conf repaired on a simulated missing-conf data dir) would close this and prevent a regression.
2. **scram-vs-md5 for upgraded installs.** The repaired `pg_hba` requires `scram-sha-256`. If a pre-Chainguard (bitnami) upgrade stored its passwords as md5 hashes, clients won't authenticate even though Postgres now starts. Worth confirming upgraded installs' credentials are scram-hashed, or handling the md5 case.

Happy to flip to approve once the entrypoint path + the md5/scram case are confirmed (a smoke test would do both).

postgresql/pg_hba.conf
```diff
@@ -0,0 +1,128 @@
+# PostgreSQL Client Authentication Configuration File
```
danielhoherd: I am concerned about us introducing these files. Do we really want ap-vendor to become the source of truth for these files during an upgrade? Having these files here means we are taking on new responsibilities that were previously taken care of by bitnami / chainguard. Are we willing to put in the engineering hours to maintain these files if something in them creates problems for customers, especially for a feature that is not intended to be used in production?

danielhoherd: It looks like there is a `/usr/share/postgresql17/pg_hba.conf.sample` file we can use for this:

```
$ docker run --entrypoint=/bin/bash cgr.dev/astronomer.io/ap-postgres:17-dev -c "ls -1 /usr/share/postgresql17/ | grep conf"
pg_hba.conf.sample
pg_ident.conf.sample
pg_service.conf.sample
postgresql.conf.sample
```

I'd much rather us copy those than store these files in our own repo.

pgvishnuram: @danielhoherd we should descope this - this is not for production setup and we should not encourage incluster for production setup

rishkarajgi: Is this required to fix the underlying issue?

postgresql/docker-entrypoint-wrapper.sh
Shubham-astro: this has been derived from chainguard itself: https://console.chainguard.dev/org/astronomer.io/images/organization/image/ap-postgres/specifications

postgresql/pg_hba.conf
Shubham-astro: @rishkarajgi yes, this is a required change, if we want the upgrade to be stable. 

@danielhoherd I checked this, they contain unsubstituted initdb template placeholders (@authmethodlocal@, @authmethodhost@, etc. — confirmed by pulling the raw file from cgr.dev/astronomer.io/ap-postgres:17-dev), so referencing them as-is would write a broken pg_hba.conf and fail to start. There's also a rule in our shipped config (host all all all scram-sha-256) that isn't in the sample template at all — it's added by the image's own entrypoint after initdb runs, and we don't have source access to that logic to reproduce it reliably. So deriving this from the sample would mean reimplementing both initdb's substitution and reverse-engineering undocumented entrypoint behavior — more surface area to get wrong than a verified copy of real, working output.

postgresql/pg_hba.conf
Shubham-astro: @rishkarajgi , I checked a live cluster's pg_authid, and this one's postgres role is already SCRAM-hashed, but that's just one cluster's history, not a guarantee across the fleet (anyone whose password predates PG14's SCRAM default could still be md5-hashed). Since md5 as the hba method auto-upgrades to real SCRAM auth when that's what's stored, and only adds coverage for legacy MD5 verifiers, I've swapped that line to md5, zero cost either way and it removes the risk.

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,73 @@
...
+# password predates Postgres's SCRAM default and was never rotated - see
+# PINF-347 review discussion.
+
+data_dir="${PGDATA:-/bitnami/postgresql/data}"
+
+if [ -f "$data_dir/PG_VERSION" ] && { [ ! -f "$data_dir/postgresql.conf" ] || [ ! -f "$data_dir/pg_hba.conf" ]; }; then
+  scratch_dir="/tmp/pg-repair-$$"
+
+  if mkdir -p "$scratch_dir" 2>/dev/null \
+    && initdb --auth-local=trust --auth-host=trust --username=postgres --no-sync "$scratch_dir" >/dev/null 2>&1; then
```
rishkarajgi: This assumes `initdb --auth-host=trust` only ever writes loopback-scoped host rules (127.0.0.1/32, ::1/128) -- which is standard, documented Postgres behavior, and matches the comment above about needing a separate hand-maintained remote rule -- but I couldn't run initdb here to confirm it empirically, and if that assumption is wrong (a broader host rule generated instead) the appended `md5` line below would never be reached and every remote connection would authenticate via trust. A boot smoke test would settle this the same way it'd settle the entrypoint-path question below.

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,73 @@
...
+      echo "host all all all md5" >> "$data_dir/pg_hba.conf"
+    fi
+  else
+    echo "docker-entrypoint-wrapper.sh: scratch initdb repair failed; leaving config as-is" >&2
+  fi
+
+  rm -rf "$scratch_dir"
+fi
+
+exec /usr/bin/docker-entrypoint.sh "$@"
```
rishkarajgi, line 68: Same ask as the 20th: this hands off to a hardcoded path that isn't verifiable from this repo, and the Chainguard console link in the earlier thread is documentation, not a test -- if the path is wrong every Postgres pod fails to start. Also worth confirming while at it: does `/usr/bin/docker-entrypoint.sh`'s own first-init path invoke initdb with the same `--auth-local`/`--auth-host` flags this wrapper uses, so the generated config actually matches what a fresh install would produce? A single boot test (repair path + a real client connection) would cover both.

postgresql/endorignore
```diff
@@ -0,0 +1,2 @@
+CVE-2026-14456
```
rishkarajgi: What are these two CVEs, and is there a ticket recording why they're suppressed here? Worth a comment or a linked issue so this doesn't read as a blanket ignore to the next person who finds it.

rishkarajgi (commented):
This has changed shape substantially since the last round, and it's a real improvement: instead of shipping and maintaining a static, hand-copied postgresql.conf/pg_hba.conf (the version reviewed on the 19th/20th), the wrapper now runs a scratch `initdb` with this same image's real binary and lifts the two generated files out of a throwaway directory. That resolves danielhoherd's "do we want to own these files" objection directly -- ap-vendor isn't the source of truth for Postgres config anymore, it's just relaying what the image's own initdb produces -- and it makes the Copilot/depthfirst-app findings on the old static pg_hba.conf/postgresql.conf moot: those files are explicitly commented out of the COPY and kept only for reference/rollback, so they're not live in the running image. The PGDATA default is also fixed to match the real upgrade path (`/bitnami/postgresql/data`), and the scram/md5 question from the 20th is resolved cleanly (`md5` accepts both SCRAM and legacy MD5 verifiers, zero downside).

What's still open, though, is the same category of risk the last review was about, just relocated: the entrypoint-path confirmation is still only a link to Chainguard's console page, not an actual verified boot, and the new approach adds a second unverifiable-from-this-repo assumption on top of it (inline). Both would be closed by the same boot smoke test that was asked for on the 20th, which still doesn't exist anywhere in this diff -- for the image every internal Postgres runs on, I'd still want that before merge rather than a documented assumption, same reasoning as the original request.

Two smaller things, also inline: the CVE suppressions in `endorignore` have no visible justification or linked ticket, and it's worth being explicit in the PR description (not just a Dockerfile comment) that the 847+128 lines of now-dead reference config are intentional rollback insurance, not a leftover -- a future reviewer skimming the file list won't know that without reading the comment first.

This is a secrets/auth-adjacent platform change (Postgres client authentication for every internal install), so it needs a security review in addition to the usual approval.

postgresql/pg_hba.conf
Shubham-astro: This is resolved, we are not coping the sample and replacing values in it. @danielhoherd

postgresql/endorignore
Shubham-astro: removed

postgresql/docker-entrypoint-wrapper.sh
Shubham-astro: This path is referenced from chainguard documentation.

danielhoherd: FTR:

```
$ docker run -ti --entrypoint=/bin/bash cgr.dev/astronomer.io/ap-postgres:17-dev -c 'head /usr/bin/docker-entrypoint.sh'
#!/usr/bin/env bash
set -Eeo pipefail
# TODO swap to -Eeuo pipefail above (after handling all potentially-unset variables)

# usage: file_env VAR [DEFAULT]
#    ie: file_env 'XYZ_DB_PASSWORD' 'example'
# (will allow for "$XYZ_DB_PASSWORD_FILE" to fill in the value of
#  "$XYZ_DB_PASSWORD" from a file, especially for Docker's secrets feature)
file_env() {
	local var="$1"
```

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,69 @@
...
+        gid="$(id -g)"
+        printf 'postgres:x:%s:%s:PostgreSQL:%s:/bin/false\n' "$uid" "$gid" "$data_dir" > "$NSS_WRAPPER_PASSWD"
+        printf 'postgres:x:%s:\n' "$gid" > "$NSS_WRAPPER_GROUP"
+        break
+      fi
+    done
+  fi
+
+  if mkdir -p "$scratch_dir" 2>/dev/null; then
+    if initdb_output="$(initdb --auth-local=trust --auth-host=trust --username=postgres --no-sync "$scratch_dir" 2>&1)"; then
```
danielhoherd: **Repaired local/loopback auth is weaker than what this image's own fresh-init path produces.**

Checked the real entrypoint directly (`docker run --entrypoint=/bin/bash cgr.dev/astronomer.io/ap-postgres:17-dev`, then `cat /usr/bin/docker-entrypoint.sh`):

- `docker_init_database_dir()` calls plain `initdb --username="$POSTGRES_USER" --pwfile=<(...) $POSTGRES_INITDB_ARGS "$@"` — no `--auth-local`/`--auth-host` at all. A password is always supplied (`docker_verify_minimum_env` enforces that unless `POSTGRES_HOST_AUTH_METHOD=trust` is explicitly chosen), and the comment directly above `pg_setup_hba_conf()` notes PG14+ changed initdb's default away from trust specifically to stop shipping passwordless auth by default. So on PG17 with a password present and no explicit override, initdb's default for local + host loopback is `scram-sha-256`.
- `pg_setup_hba_conf()` doesn't touch local/host at all — it only appends one `host all all all $POSTGRES_HOST_AUTH_METHOD` line, derived from `postgres -C password_encryption` (effectively scram-sha-256).

So a genuinely fresh install of this same image ends up with `scram-sha-256` on local socket + 127.0.0.1/::1. This wrapper's scratch `initdb --auth-local=trust --auth-host=trust` instead bakes `trust` into the copied `pg_hba.conf`'s local + loopback rules on every repaired install — strictly weaker than what this exact image's own fresh-init path gives you, so anyone who can reach the pod's socket or loopback can authenticate as any role, including the superuser, without a password.

The `md5` choice for the remote rule below is fine — `md5` in pg_hba transparently upgrades to a real SCRAM exchange when the stored verifier is SCRAM, so it's a superset covering legacy MD5-hashed passwords too, not a downgrade. That reasoning doesn't extend to local/host though: a repaired instance authenticates the same stored password either way, so there's no compatibility reason to weaken those to `trust` — this looks like it was just for scratch-initdb convenience (skips plumbing `POSTGRES_PASSWORD`/`password_encryption` into the throwaway instance) rather than a deliberate tradeoff.

```suggestion
    if initdb_output="$(initdb --auth-local=scram-sha-256 --auth-host=scram-sha-256 --username=postgres --no-sync "$scratch_dir" 2>&1)"; then
```

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,69 @@
...
+# image's own `initdb` against a throwaway scratch dir (can't target the real,
+# already-populated $PGDATA - initdb refuses a non-empty directory) and copy the
+# two generated files into place. That way the repair always matches whatever the
+# running image's initdb actually produces, instead of a static copy we'd have to maintain.
+
+data_dir="${PGDATA:-/bitnami/postgresql/data}"
+
+if [ -f "$data_dir/PG_VERSION" ] && { [ ! -f "$data_dir/postgresql.conf" ] || [ ! -f "$data_dir/pg_hba.conf" ]; }; then
+  scratch_dir="/tmp/pg-repair-scratch"
+  rm -rf "$scratch_dir"
```
danielhoherd: **Predictable scratch dir under shared `/tmp` — same TOCTOU depthfirst-app flagged on the earlier version of this script, doesn't look like it carried forward into this rewrite.**

Fixed path + `rm -rf` then (later, line 34) `mkdir -p` isn't exclusive/atomic: any other process that can write to this pod's `/tmp` can pre-create `/tmp/pg-repair-scratch` (e.g. as a symlink) or race the window between the `rm -rf` and the `mkdir -p`, and have its content picked up as the "generated" config that gets copied into `$PGDATA`. `mktemp -d` creates the directory atomically with `0700` perms and a random suffix, closing that window.

```suggestion
  scratch_dir="$(mktemp -d /tmp/pg-repair-scratch.XXXXXX 2>/dev/null)" || scratch_dir=""
```

(This needs a matching change at the `mkdir -p "$scratch_dir"` check below — left as a separate suggestion on that line, since `mktemp -d` already creates the directory so that line just needs to check `scratch_dir` is non-empty instead of trying to create it again.)

postgresql/docker-entrypoint-wrapper.sh
```diff
@@ -0,0 +1,69 @@
...
+        export LD_PRELOAD="$wrapper" NSS_WRAPPER_PASSWD NSS_WRAPPER_GROUP
+        gid="$(id -g)"
+        printf 'postgres:x:%s:%s:PostgreSQL:%s:/bin/false\n' "$uid" "$gid" "$data_dir" > "$NSS_WRAPPER_PASSWD"
+        printf 'postgres:x:%s:\n' "$gid" > "$NSS_WRAPPER_GROUP"
+        break
+      fi
+    done
+  fi
+
+  if mkdir -p "$scratch_dir" 2>/dev/null; then
```
danielhoherd: Companion change for the `mktemp -d` suggestion above — `scratch_dir` is already created by `mktemp -d` at that point, so this just needs to check it came back non-empty rather than trying to `mkdir` it again.

```suggestion
  if [ -n "$scratch_dir" ]; then
```

danielhoherd (commented):
Pulled the real `/usr/bin/docker-entrypoint.sh` off `cgr.dev/astronomer.io/ap-postgres:17-dev` to check two things against the actual source rather than assumption. Two findings, both with suggested fixes inline. (Also: understood smoke-testing a hardened image isn't possible today given the open issue for that — not asking for one here.)

postgresql/docker-entrypoint-wrapper.sh
Shubham-astro: @danielhoherd  applied your suggestion, and upon upgrade the upgrade failed, Checked this against the actual initdb.c source (PG17): when no --auth/--auth-local/--auth-host is given, the fallback is unconditionally trust regardless of PG version — the PG14+ default change you're citing is password_encryption's default, which only affects the separate pg_setup_hba_conf() remote-line logic (a different function, already handled — see the md5 line below), not docker_init_database_dir()'s bare initdb call. Also confirmed empirically: a genuinely fresh install of this exact image produces trust for local/loopback today. Keeping trust here so repaired installs match what a fresh install of this image actually does.

danielhoherd (approved):
I've dug in a bit and I think this is good to merge.

